PR #2, opened 2017-08-04
[WIP] Initial draft.

smola:


bzz:
Nice API!

If one reverse-engineer main idea of this proposal - it is

```scala
val filsRDD = GitLoad.clone(tmpDir, urlRDD).flatMapGit { git =>
  ....
}

val filesDF = sc.createDataFrame()

filesDF
 .withColumn("language", detectLanguage(col("file.path"), col("blob.content")))
 .withColumn("uast", uast(col("language"), col("blob.content")))
```

*Major missing parts*:
 - API for next steps, that include calling Python code over UAST (token extraction)
 - for current steps: way to be exposed to Python
 - HDFS access to .siva files
 - cleanup of local FS after traversing all files in a repo
 - optimization: this impl has additional serialization, save all fils on-disk before filtering ref 
 - optimization: looks like LangID using Enry and UAST parsing in Bblfsh are not re-use same in-memory blob content representation (in src-d/berserker they do) 
 - error handling

PR #17, opened 2017-08-31
First approach to git datasource - do not merge, for reference only

ajnavarro:


erizocosmico:
Since this is for reference only and it's not gonna be merged, should we review it as well?

ajnavarro:
@erizocosmico not needed. Thanks.

PR #19, opened 2017-09-01
Repository provider implementation

ajnavarro:
Added a way to obtain a repository instance using a local path and a hadoop configuration.

This PR contains two providers:
- SivaRDDProvider provides always the same RDD instance per remote siva folder.
- RepositoryProvider. Intended to be used inside Spark transformations, it downloads the siva file, unpack it, deletes the unpacked siva file, and create a repository instance pointing to the local folder. If the instance already exists it doesn't download and unpack again the siva file.


ajnavarro:
# Merge https://github.com/src-d/spark-api/pull/18 first

src/main/scala/tech/sourced/api/provider/RepositoryProvider.scala
```diff
@@ -0,0 +1,91 @@
...
+      // FileUtils.deleteQuietly(r.getDirectory)
+      case None =>
+    }
+  }
+}
+
+object RepositoryProvider {
+  var provider: RepositoryProvider = _
+
+  def apply(localPath: String): RepositoryProvider = {
```
erizocosmico: if this is a singleton, what about moving everything to the `RepositoryProvider` object instead of having the class and manually manage the singleton?

Also, if you do `RepositoryProvider("foo")` and then `RepositoryProvider("bar")` what you get is a repository provider with `"foo"` as localPath, which is a bit misleading.

src/main/scala/tech/sourced/api/provider/SivaRDDProvider.scala
```diff
@@ -0,0 +1,36 @@
...
+        rdd
+      case Some(rdd) => rdd
+    }
+  }
+}
+
+object SivaRDDProvider {
+  var provider: SivaRDDProvider = _
+
+  def apply(sc: SparkContext): SivaRDDProvider = {
```
erizocosmico: same as with `RepositoryProvider`

src/main/scala/tech/sourced/api/provider/RepositoryProvider.scala
ajnavarro: Yep, you're right. I did this as it is because in a near future maybe we are going need a repository provider per path.

In the way that is now we should throw an error if a provider instance exists but not with the same `localPath` provided. What do you think?

erizocosmico: Ok, makes sense. Yeah, maybe we should throw an error in that case :+1:

src/main/scala/tech/sourced/api/provider/SivaRDDProvider.scala
ajnavarro: In this case is not necessary. Spark throws an error if we try to create more than one SparkContext instance.

erizocosmico: Then the class is not necessary, as it's always going to be used as a singleton, right?

ajnavarro:
> It would be nice to have a little ScalaDoc for classes/objects though.
I was in a rush to have something working before the demo. I will add documentation for all the classes after have something working, if you don't mind.

PR #20, opened 2017-09-04
	RepositoryIterator and ReferenceIterator implementations

ajnavarro:
Using a base abstract class RootedRepoIterator, we add two implementations, one of them to iterate repository metadata (repository id, urls, is fork) and references metadata (repository_id, name, hash).

With this RootedRepoIterator we should be able to implement CommitIterator and BlobIterator too.

Filter logic must be implemented before start with BlobIterator.

- Split test logic into Traits to be able to use them in all the Specs.
- Added a BaseRootedRepoIterator trait with a helper to test iterators more easly.

### DIFF
--- a/src/main/scala/tech/sourced/api/util/GitUrlsParser.scala
+++ b/src/main/scala/tech/sourced/api/util/GitUrlsParser.scala
@@ -1,0 +1,22 @@
+package tech.sourced.api.util
+
+import java.net.{URI, URISyntaxException}
+
+import scala.util.matching.Regex
+
+object GitUrlsParser {
+  val isGit: Regex = """(.+)\\@(.+):(.+)\\.git""".r
+
+  def getIdFromUrls(urls: Array[String]): String = {
+    urls.flatMap({
+      case isGit(_, host, path, _*) => Some(host + path)
+      case s => try {
+        val u: URI = new URI(s)
+        Some(u.getHost + u.getPath)
+      } catch {
+        case _: URISyntaxException => None
+      }
+    }).distinct.min
+  }
+
+}
--- a/src/main/scala/tech/sourced/api/iterator/ReferenceIterator.scala
+++ b/src/main/scala/tech/sourced/api/iterator/ReferenceIterator.scala
@@ -1,0 +1,36 @@
+package tech.sourced.api.iterator
+
+import org.eclipse.jgit.lib.{ObjectId, Ref, Repository}
+
+import scala.collection.JavaConverters._
+
+class ReferenceIterator(requiredColumns: Array[String], repo: Repository)
+  extends RootedRepoIterator[Ref](requiredColumns, repo) {
+
+  override protected def loadIterator(): Iterator[Ref] =
+    repo.getAllRefs.asScala.values.toIterator
+
+  override protected def mapColumns(ref: Ref): Map[String, () => Any] = {
+    val (repoId, refName) = parseRef(ref.getName)
+    Map[String, () => Any](
+      "repository_id" -> (() => {
+        repoId
+      }),
+      "name" -> (() => {
+        refName
+      }),
+      "hash" -> (() => {
+        ObjectId.toString(ref.getObjectId)
+      })
+    )
+  }
+
+  private def parseRef(ref: String): (String, String) = {
+    val split: Array[String] = ref.split("/")
+    val uuid: String = split.last
+    val repoId: String = this.getRepositoryId(uuid).get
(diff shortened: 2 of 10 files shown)


ajnavarro:
# Merge https://github.com/src-d/spark-api/pull/19 first.

src/main/scala/tech/sourced/api/util/GitUrlsParser.scala
```diff
@@ -0,0 +1,22 @@
...
+  def getIdFromUrls(urls: Array[String]): String = {
+    urls.flatMap({
+      case isGit(_, host, path, _*) => Some(host + path)
+      case s => try {
+        val u: URI = new URI(s)
+        Some(u.getHost + u.getPath)
+      } catch {
+        case _: URISyntaxException => None
+      }
+    }).distinct.min
```
erizocosmico: Is `min` used because of something specific or just to get one of the results?

src/main/scala/tech/sourced/api/provider/RepositoryProvider.scala
```diff
@@ -0,0 +1,91 @@
...
+        r.close()
+      // TODO maybe others are using this repository instance
+      // FileUtils.deleteQuietly(r.getDirectory)
+      case None =>
+    }
+  }
+}
+
+object RepositoryProvider {
+  var provider: RepositoryProvider = _
```
erizocosmico: if this is a singleton, why not move everything to `RepositoryProvider` instead of having the class and manually manage the singleton?

Also, if you do `RepositoryProvider("foo")` and then `RepositoryProvider("bar")` what you get is a repository provider with `"foo"` as localPath, which is a bit misleading.

src/main/scala/tech/sourced/api/provider/SivaRDDProvider.scala
```diff
@@ -0,0 +1,36 @@
...
+        rdd
+      case Some(rdd) => rdd
+    }
+  }
+}
+
+object SivaRDDProvider {
+  var provider: SivaRDDProvider = _
+
+  def apply(sc: SparkContext): SivaRDDProvider = {
```
erizocosmico: same as with `RepositoryProvider`

src/main/scala/tech/sourced/api/util/GitUrlsParser.scala
ajnavarro: min === sorted.first

src/main/scala/tech/sourced/api/provider/RepositoryProvider.scala
ajnavarro: This code is from another PR. Can you comment this there? https://github.com/src-d/spark-api/pull/19

src/main/scala/tech/sourced/api/provider/SivaRDDProvider.scala
ajnavarro: This code is from another PR. Can you comment this there?: https://github.com/src-d/spark-api/pull/19

src/main/scala/tech/sourced/api/provider/RepositoryProvider.scala
erizocosmico: Sure!

src/main/scala/tech/sourced/api/provider/SivaRDDProvider.scala
erizocosmico: Done: https://github.com/src-d/spark-api/pull/19/files#r136828317

src/main/scala/tech/sourced/api/util/GitUrlsParser.scala
erizocosmico: yes, that's what I meant, do we need them sorted or do we just want the first?

ajnavarro: we want the first after sort them.

erizocosmico: Discussed IRL: sorted is needed, lgtm then

bzz: I must admit that it's not clear to me either, why sorting is needed.
Could be a good idea to document that

ajnavarro:
> Having all changes from a different PRs together make review harder

Yes, sorry about that, but at this early stage on the project, is really difficult split functionality and go forward without depending of another in-process functionalities.

PR #24, opened 2017-09-07
ColumnFilter helper to use into iterator

ajnavarro:
ColumnFilter object that is able to:

- Return true or false to check if we need to filter a specific row from the iteration
- Get a list of matching elements to be able to filter directly using them

### DIFF
--- a/src/main/scala/tech/sourced/api/util/ColumnFilter.scala
+++ b/src/main/scala/tech/sourced/api/util/ColumnFilter.scala
@@ -1,0 +1,83 @@
+package tech.sourced.api.util
+
+import org.apache.spark.sql.sources._
+
+object ColumnFilter {
+  def compileFilter(f: Filter): CompiledFilter = {
+    f match {
+      case EqualTo(attr, value) => new EqualFilter(attr, value)
+      case EqualNullSafe(attr, value) => new EqualFilter(attr, value)
+      case IsNull(attr) => new EqualFilter(attr, null)
+      case IsNotNull(attr) => NotFilter(new EqualFilter(attr, null))
+      case Not(child) => NotFilter(compileFilter(child))
+      case And(l, r) => new AndFilter(compileFilter(l), compileFilter(r))
+      case Or(l, r) => new OrFilter(compileFilter(l), compileFilter(r))
+      case _ => UnhandledFilter()
+    }
+  }
+}
+
+sealed trait CompiledFilter {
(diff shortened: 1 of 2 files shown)


ajnavarro:
@bzz can you check if this filter implementation cover all your needs for BlobsIterator please? Thanks!

src/main/scala/tech/sourced/api/util/ColumnFilter.scala
```diff
@@ -0,0 +1,75 @@
...
+
+trait CompiledFilter {
+  def eval(cols: Map[String, Any]): Option[Boolean]
+
+  def matchingCases: Map[String, Seq[Any]] =
+    this.getMatchingCases.groupBy(_._1).mapValues(_.map(_._2))
+
+  private def getMatchingCases: Array[(String, Any)] =
+    this match {
+      case ef: EqualFilter => Array((ef.name, ef.value))
```
ajnavarro: we need to have into account the `Not` modifier

ajnavarro: done.

bzz:
👍 looking into it right now

src/main/scala/tech/sourced/api/util/ColumnFilter.scala
```diff
@@ -0,0 +1,75 @@
...
+      case IsNotNull(attr) => NotFilter(new EqualFilter(attr, null))
+      case Not(child) => NotFilter(compileFilter(child))
+      case And(l, r) => new AndFilter(compileFilter(l), compileFilter(r))
+      case Or(l, r) => new OrFilter(compileFilter(l), compileFilter(r))
+      case _ => UnhandledFilter()
+    }
+  }
+}
+
+trait CompiledFilter {
```
erizocosmico: shouldn't we seal it to ensure exhaustiveness when pattern matching over this type?

bzz:
With current API the use case for getting commit hashes will look like:

```scala
  filters.flatMap { filter =>
      filter.matchingCases("hash").flatMap { hash => 
         //
      }
  }.toIterator
```

I'm still a bit confused on i.e how will the eq filter get created say, for the commits.

bzz:
One more thing - it would be really nice to have .toString() defined for CompileFilter.

Similar way to .explain(), but without the truncation - this would be of a great aid while debugging

bzz:
👍 

is there anything left, before merging this guy?

PR #25, opened 2017-09-08
BlobIterator implementation

bzz:
In commit table, we have hashes of commits, which are used to get trees -> blobs.

Current implementation does not filter:
 - repository (so same blob can happen in different repository)
 - refs

It leverages `ColumnFilters` from #24 (need to be merged first) in order to get particular commit hashes, rather then iterating all refs HEADs.

### DIFF
--- a/src/main/scala/tech/sourced/api/Implicits.scala
+++ b/src/main/scala/tech/sourced/api/Implicits.scala
@@ -32,7 +32,8 @@ object Implicits {
       Implicits.checkCols(df, "hash")
       val blobsIdsDf = df.select($"hash").distinct()
       val filesDf = Implicits.getDataSource("files", df.sparkSession)
-      filesDf.join(blobsIdsDf, filesDf("commit_hash") === df("hash")).drop($"hash")
+      val filesDfJoined = filesDf.join(blobsIdsDf, filesDf("commit_hash") === blobsIdsDf("hash")).drop($"hash")
+      df.join(filesDfJoined, df("hash") === filesDfJoined("commit_hash"))
     }
   }
 
--- a/src/main/scala/tech/sourced/api/Implicits.scala
+++ b/src/main/scala/tech/sourced/api/Implicits.scala
@@ -29,10 +29,15 @@ object Implicits {
     }
 
     def getFiles: DataFrame = {
-      Implicits.checkCols(df, "hash")
-      val commitsDf = df.drop("tree").distinct()
       val filesDf = Implicits.getDataSource("files", df.sparkSession)
-      filesDf.join(commitsDf, filesDf("commit_hash") === commitsDf("hash")).drop($"hash")
+
+      if (df.schema.fieldNames.contains("hash")) {
+        val commitsDf = df.drop("tree").distinct()
+        filesDf.join(commitsDf, filesDf("commit_hash") === commitsDf("hash")).drop($"hash")
+      } else {
+        Implicits.checkCols(df, "reference_name")
+        filesDf
+      }
     }
   }
 
--- a/src/main/scala/tech/sourced/api/Implicits.scala
+++ b/src/main/scala/tech/sourced/api/Implicits.scala
@@ -29,10 +29,10 @@ object Implicits {
     }
 
     def getFiles: DataFrame = {
-      Implicits.checkCols(df, "tree")
-      val blobsIdsDf = df.select($"blobs").distinct()
+      Implicits.checkCols(df, "hash")
+      val blobsIdsDf = df.select($"hash").distinct()
       val filesDf = Implicits.getDataSource("files", df.sparkSession)
-      filesDf.join(blobsIdsDf, array_contains(blobsIdsDf("blobs"), filesDf("file_hash"))).drop($"blobs")
+      filesDf.join(blobsIdsDf, filesDf("commit_hash") === df("hash")).drop($"hash")
     }
   }
 
--- a/src/main/scala/tech/sourced/api/Implicits.scala
+++ b/src/main/scala/tech/sourced/api/Implicits.scala
@@ -30,10 +30,9 @@ object Implicits {
 
     def getFiles: DataFrame = {
       Implicits.checkCols(df, "hash")
-      val blobsIdsDf = df.select($"hash").distinct()
+      val commitsDf = df.drop("tree").distinct()
       val filesDf = Implicits.getDataSource("files", df.sparkSession)
-      val filesDfJoined = filesDf.join(blobsIdsDf, filesDf("commit_hash") === blobsIdsDf("hash")).drop($"hash")
-      df.join(filesDfJoined, df("hash") === filesDfJoined("commit_hash"))
+      filesDf.join(commitsDf, filesDf("commit_hash") === commitsDf("hash")).drop($"hash")
     }
   }
 
--- a/src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
+++ b/src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
@@ -39,9 +39,9 @@ class BlobIterator(requiredColumns: Array[String], repo: Repository, filters: Ar
       filtered
     } else {
       val refs = new Git(repo).branchList().call().asScala.filter(!_.isSymbolic)
-      println(s"Iterating all ${refs.size} refs")
+      log.warn(s"Iterating all ${refs.size} refs")
       refs.toIterator.flatMap { ref =>
-        println(s" $ref")
+        log.warn(s" $ref")
         JGitBlobIterator(getTreeWalk(ref.getObjectId), log)
       }
     }
--- a/src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
+++ b/src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
@@ -1,0 +1,114 @@
+package tech.sourced.api.iterator
+
+import org.apache.log4j.Logger
+import org.eclipse.jgit.diff.RawText
+import org.eclipse.jgit.lib.{ObjectId, ObjectReader, Repository}
+import org.eclipse.jgit.revwalk.RevWalk
+import org.eclipse.jgit.treewalk.TreeWalk
+import tech.sourced.api.util.CompiledFilter
+
+/**
+  * Blob iterator: returns all blobs from the filtered commits
+  *
+  * @param requiredColumns
+  * @param repo
+  * @param filters
+  */
+class BlobIterator(requiredColumns: Array[String], repo: Repository, filters: Array[CompiledFilter])
+  extends RootedRepoIterator[TreeWalk](requiredColumns, repo) {
+
+  val log = Logger.getLogger(this.getClass.getSimpleName)
+  val readMaxBytes = 20 * 1024 * 1024
+
+  override protected def loadIterator(): Iterator[TreeWalk] = {
+    filters.flatMap { filter =>
+      filter.matchingCases.getOrElse("hash", Seq()).flatMap { hash =>
+        val commitId = ObjectId.fromString(hash.asInstanceOf[String])
+        if (repo.hasObject(commitId)) {
+          val revWalk = new RevWalk(repo)
+          val revCommit = revWalk.parseCommit(commitId)
+          revWalk.close()
+
+          val treeWalk = new TreeWalk(repo)
+          treeWalk.setRecursive(true)
+          treeWalk.addTree(revCommit.getTree)
+
+          JGitBlobIterator(treeWalk, log)
+        } else {
+          Seq()
+        }
+      }
+    }
+  }.toIterator
+
+  override protected def mapColumns(tree: TreeWalk): Map[String, () => Any] = {
+    val content = readFile(tree.getObjectId(0), tree.getObjectReader)
+    Map[String, () => Any](
+      "file_hash" -> (() => tree.getObjectId(0)),
+      "content" -> (() => content),
+      "is_binary" -> (() => RawText.isBinary(content)),
+      "path" -> (() => tree.getPathString)
+    )
+  }
+
+  /**
+    * Read max N bytes of the given blob
+    *
+    * @param objId
+    * @param reader
+    * @param max maximum number of bytes to read in memory
+    * @return
+    */
+  def readFile(objId: ObjectId, reader: ObjectReader, max: Integer = readMaxBytes): Array[Byte] = {
+    val obj = reader.open(objId)
+    val data = if (obj.isLarge) {
+      val buf = Array.ofDim[Byte](max)
+      val is = obj.openStream()
+      is.read(buf)
+      is.close()
+      buf
+    } else {
+      obj.getBytes
+    }
+    reader.close()
+    data
+  }
+
+}
+
+
+class JGitBlobIterator[T <: TreeWalk](treeWalk: T, log: Logger) extends Iterator[T] {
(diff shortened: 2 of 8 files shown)


bzz:
This is not supposed to be complete implementation yet: i.e it does nothing at all, in case there were no filters for commits.

One thing that seems reasonable in such cases - iterate all (non-remote, non-filtered) refs HEADS.

Feedback is very welcome.

\cc @ajnavarro @mcarmonaa

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -0,0 +1,113 @@
...
+          Seq()
+        }
+      }
+    }
+  }.toIterator
+
+  override protected def mapColumns(tree: TreeWalk): Map[String, () => Any] = {
+    val content = readFile(tree.getObjectId(0), tree.getObjectReader)
+    Map[String, () => Any](
+      "file_hash" -> (() => tree.getObjectId(0)),
```
ajnavarro: you miss the commit hash maybe?

bzz:
Join in in Implicits commits and files on `"commit_hash" === "hash"` does not produce a `eq` filter (only `is_not_null(commit_hash)`) so a new case was added that iterates blobs in HEADs of all refs

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -0,0 +1,155 @@
...
+  * Blob iterator: returns all blobs from the filtered commits
+  *
+  * @param requiredColumns
+  * @param repo
+  * @param filters
+  */
+class BlobIterator(requiredColumns: Array[String], repo: Repository, filters: Array[CompiledFilter])
+  extends RootedRepoIterator[CommitTree](requiredColumns, repo) {
+
+  val log = Logger.getLogger(this.getClass.getSimpleName)
```
ajnavarro: You should use `org.apache.spark.internal.Logging`. See an example in `org.apache.spark.sql.execution.datasources.jdbc.JDBCRDD`.

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -0,0 +1,155 @@
...
+
+  override protected def loadIterator(): Iterator[CommitTree] = {
+    val filtered = filters.toIterator.flatMap { filter =>
+      filter.matchingCases.getOrElse("hash", Seq()).flatMap { hash =>
```
ajnavarro: You can do here `filter.matchingCases("hash).flatMap...` directly.

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -0,0 +1,155 @@
...
+  * @param repo
+  * @param filters
+  */
+class BlobIterator(requiredColumns: Array[String], repo: Repository, filters: Array[CompiledFilter])
+  extends RootedRepoIterator[CommitTree](requiredColumns, repo) {
+
+  val log = Logger.getLogger(this.getClass.getSimpleName)
+
+  override protected def loadIterator(): Iterator[CommitTree] = {
+    val filtered = filters.toIterator.flatMap { filter =>
```
ajnavarro: Right now I am not totally sure if moving `toIterator` after the flatMap can improve performance here. I usually follow the rule of "only apply transformations when is totally necessary" and here you can do a `flatMap` without transform filters to an iterator.

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -0,0 +1,155 @@
...
+        }
+      }
+    }
+
+    if (filtered.hasNext) {
+      filtered
+    } else {
+      val refs = new Git(repo).branchList().call().asScala.filter(!_.isSymbolic)
+      log.warn(s"Iterating all ${refs.size} refs")
+      refs.toIterator.flatMap { ref =>
```
ajnavarro: the same as commented before.

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -0,0 +1,155 @@
...
+          Seq()
+        }
+      }
+    }
+
+    if (filtered.hasNext) {
+      filtered
+    } else {
+      val refs = new Git(repo).branchList().call().asScala.filter(!_.isSymbolic)
+      log.warn(s"Iterating all ${refs.size} refs")
```
ajnavarro: this is a warning or a debug?

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -0,0 +1,155 @@
...
+      refs.toIterator.flatMap { ref =>
+        log.warn(s" $ref")
```
ajnavarro: the same as commented before

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -0,0 +1,155 @@
...
+        JGitBlobIterator(getTreeWalk(ref.getObjectId), log)
+      }
+    }
+  }
+
+  override protected def mapColumns(commitTree: CommitTree): Map[String, () => Any] = {
+    val content = BlobIterator.readFile(commitTree.tree.getObjectId(0), commitTree.tree.getObjectReader)
+    Map[String, () => Any](
+      "file_hash" -> (() => commitTree.tree.getObjectId(0).name),
+      "content" -> (() => content),
```
ajnavarro: Should we return the content of a binary file?

src/test/scala/tech/sourced/api/DefaultSourceSpec.scala
```diff
@@ -24,13 +24,15 @@ class DefaultSourceSpec extends FlatSpec with Matchers with BaseSivaSpec with Ba
 
     commitsDf.show()
 
-    // TODO files not implemented yet
-    //    val filesDf = ss.read.format("tech.sourced.api")
-    //      .option("table", "files")
-    //      .load(resourcePath)
-    //
-    //    filesDf.withColumn("content string", filesDf("content").cast(StringType)).show()
+    println("Files/blobs (without commit hash filtered) at HEAD or every ref:\n")
```
ajnavarro: Use Scalatest method `info`:  `info("Files/blobs (without commit hash filtered) at HEAD or every ref:\n")`

src/test/scala/tech/sourced/api/DefaultSourceSpec.scala
```diff
@@ -44,8 +46,19 @@ class DefaultSourceSpec extends FlatSpec with Matchers with BaseSivaSpec with Ba
     val reposDf = spark.getRepositories
       .filter($"id" === "github.com/mawag/faq-xiyoulinux" || $"id" === "github.com/xiyou-linuxer/faq-xiyoulinux")
     val refsDf = reposDf.getReferences.filter($"name".equalTo("refs/heads/HEAD"))
-    val commitsDf = refsDf.getCommits.select("repository_id", "reference_name", "message")
 
-    commitsDf.show()
+    val commitsDf = refsDf.getCommits.select("repository_id", "reference_name", "message", "hash")
+    //commitsDf.show()
```
ajnavarro: remove commented code

src/test/scala/tech/sourced/api/DefaultSourceSpec.scala
```diff
@@ -44,8 +46,19 @@ class DefaultSourceSpec extends FlatSpec with Matchers with BaseSivaSpec with Ba
...
+
+    println("Files/blobs with commit hashes:\n")
```
ajnavarro: use info

src/test/scala/tech/sourced/api/DefaultSourceSpec.scala
```diff
@@ -44,8 +46,19 @@ class DefaultSourceSpec extends FlatSpec with Matchers with BaseSivaSpec with Ba
...
+    val filesDf = refsDf.getCommits.getFiles.select("repository_id", "reference_name", "path", "commit_hash", "file_hash")
+    filesDf.explain(true)
+    filesDf.show()
+
+    val cnt = filesDf.count()
+    println(s"Total $cnt rows")
```
ajnavarro: use info

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
bzz: I totally see what you mean here and agree!

But the reason to convert to iterator is different - [later in code](https://github.com/src-d/spark-api/pull/25/files#diff-21e2920512108a39f3f59b247606ff7dR39) we cover the case when there are no filters, and that's when we iterate trees at HEADs of all the references.

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
bzz: Here, it is used for type conversion  - we need to get `Iterator[CommitTree]` so, in any case we need `.toIterator`, it's just the question when to do that - either here, or on the results of `.flatMap`.

But your point is valid.

bzz:
Reviews addressed, rebased on latest master. 

Going to push test now.

ajnavarro:
@bzz You should do a rebase -i and remove dd2f819 from the history. the correct one is 623560dc453cf81e77c3d1cbf9e4e2443f573820

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
ajnavarro: I see, but you can do `val filtered = filters.flatMap{...} toIterator` instead of `val filtered = filters.toIterator.flatMap{...}`

src/main/scala/tech/sourced/api/util/ColumnFilter.scala
```diff
@@ -17,7 +17,7 @@ object ColumnFilter {
   }
 }
 
-sealed trait CompiledFilter {
+trait CompiledFilter {
```
erizocosmico: why is the trait no longer sealed?

src/main/scala/tech/sourced/api/Implicits.scala
```diff
@@ -32,7 +32,8 @@ object Implicits {
       Implicits.checkCols(df, "hash")
       val blobsIdsDf = df.select($"hash").distinct()
       val filesDf = Implicits.getDataSource("files", df.sparkSession)
-      filesDf.join(blobsIdsDf, filesDf("commit_hash") === df("hash")).drop($"hash")
+      val filesDfJoined = filesDf.join(blobsIdsDf, filesDf("commit_hash") === blobsIdsDf("hash")).drop($"hash")
+      df.join(filesDfJoined, df("hash") === filesDfJoined("commit_hash"))
```
ajnavarro: I think you are joining one time more than necessary. Check this code, maybe I'm wrong.

src/main/scala/tech/sourced/api/util/ColumnFilter.scala
ajnavarro: Because the commit is not correct: https://github.com/src-d/spark-api/pull/25#issuecomment-329474015

bzz:
@ajnavarro @erizocosmico sorry, did not push the rebase before. Fixed now.

bzz:
Test are fixed in #28 

@ajnavarro on https://github.com/src-d/spark-api/pull/25#discussion_r138887872

2 joins were added, because:
 - `files.commit_hash` join \w only _uniq_ `commits.hash`, resulting table has `Schema.files` structure
 -  result is then joined with commits again, to get more fields like `repository_id`, etc

Second join can of course be manually done by the client, if we decide that's desirable.  

But rationale for having it here were client expectations on API:

```scala
spark.getRepositories.filter($"id" === "github.com/mawag/faq-xiyoulinux")
     .getReferences.filter($"name".equalTo("refs/heads/HEAD"))
     .getCommits
     .getFiles
     .select("repository_id", "reference_name", "path", "commit_hash", "file_hash")
```

Ideally, for our use-case clients would have `repository_url` here, which is not possible now, without one more join. So that could be an argument to remove extra join in files as well.

What do you think?

For me, something like below would be the best

```
spark.getRepositories.filter($"id" === "github.com/mawag/faq-xiyoulinux")
     .getReferences.filter($"name".equalTo("refs/heads/HEAD"))
     .getFiles
     .select("repository_url", "reference_name", "path", "commit_hash", "file_hash", "file_content")
```

where `repository_url` is already a main endpoint of original repository

ajnavarro:
> For me, something like below would be the best
   spark.getRepositories.filter($"id" === "github.com/mawag/faq-xiyoulinux")
     .getReferences.filter($"name".equalTo("refs/heads/HEAD"))
     .getFiles
     .select("repository_url", "reference_name", "path", "commit_hash", "file_hash", "file_content")

This is a really nice approach. We should implement it in the near future.

Regarding the two joins, you can remove the previous `df.select("hash")` statement and do something:

```scala
Implicits.checkCols(df, "hash") // checking if this datasource is a commits one
val uniqDf = df.distinct() // get unique commits
val filesDf = Implicits.getDataSource("files", df.sparkSession) // get the files datasource
filesDf.join(df, filesDf("commit_hash") === uniqDf("hash")).drop($"hash") // join by commit hash and drop the hash duplicated column (hash == commit_hash)
```
Joins are really expensive and we need to avoid them as much as possible.

bzz:
> Joins are really expensive

Addressed in 653401237b45bb66f3f891c9b5148be5c963b9d1


> This is a really nice approach. We should implement it in the near future.

Reference filter added in 3bdc052a0c7316af3f76da4a297a23a2749139bc

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
bzz: I see what you mean

> But is not important

I believe it's actually very important in this particular case, as we return JGitIterator. Here is why:
 1. `filters.flatMap{...} toIterator` will process full collection first, and then return an iterator
 2. `filters.toIterator.flatMap{...}` will return an iterator, that is ready to process the first element

And I also have spent few hours yesterday hunting 🐛 when accidentally switched between those two approaches :/

AFAIK this might not be the case, if we were only dealing with collections in memory, but in our case a stateful iterator with IO is involved.

bzz:
@ajnavarro I'm really sorry, but I do not understand what you mean by

> I think this is not the best way tho do this

I was under impression that we are not in some kind of competition for the "best way of using collection API", but rather want to get blobs and release Spark API to the users.

Please let me know if you see bugs in some corner cases of the current implementation.

bzz:
Suggestions from review applied in 03d89d0

src/test/scala/tech/sourced/api/iterator/BlobIteratorSpec.scala
```diff
@@ -0,0 +1,90 @@
+package tech.sourced.api.iterator
+
+import java.nio.charset.StandardCharsets
+
+import org.scalatest.FlatSpec
+import tech.sourced.api.util.{CompiledFilter, EqualFilter}
+
+class BlogIteratorSpec extends FlatSpec with BaseRootedRepoIterator {
```
ajnavarro: The last thing s/Blog/Blob

bzz:
CI passes now, \w fix from #37.

Merging if there is no further discussion

ajnavarro:
LGTM

PR #27, opened 2017-09-13
Add assembly plugin

ajnavarro:
- With this plugin we can generate a fatjar suitable to be sent to a Spark cluster.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,12 +11,14 @@ lazy val root = (project in file(".")).
     )),
     name := "spark-api",
     libraryDependencies += scalaTest % Test,
-    libraryDependencies += sparkSql % Compile,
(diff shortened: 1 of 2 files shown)


erizocosmico:
Should we remove the classifiers as we had to do with enry java so it was usable in spark?

ajnavarro:
I actually test the generated jar in the demo, so with the actual classifiers is working well. I don't know if I'm missing something, let me know.

bzz:
@erizocosmico good catch. 

With classifiers, it will not be possible to add this jar from directly from remote maven repository using `--packages`, which is optional, but nice to have feature.

Through the tests most probably @ajnavarro used `--jars` or equivalent in code, which works well as .jar are in local FS.

PR #26, opened 2017-09-10
[Bug] - don't merge - highlight odd behavior in SivaRDDProvider

bzz:
This is a 🐛  report - it's not intended to be merged, it just includes the code that helps to expose it.

While struggling with `MissingObjectException` in #25 I noticed that if a `println` statement is added to BlobIterator - it makes exception go away, which seems to be a sign of some race conditions or copy delay.

This PR adds visibility to the .siva file copy process.

While running a **single test** `DefaultSourceSpec. "Additional methods" should "work correctly"` it
 - always prints
   ```
   Copy file:~/src-d/spark-api/target/scala-2.11/test-classes/siva-files/05893125684f2d3943cd84a7ab2b75e53668fba1.siva to /tmp/siva-files/05893125684f2d3943cd84a7ab2b75e53668fba1.siva
   Copy file:~/src-d/spark-api/target/scala-2.11/test-classes/siva-files/05893125684f2d3943cd84a7ab2b75e53668fba1.siva to /tmp/siva-files/05893125684f2d3943cd84a7ab2b75e53668fba1.siva
   Copy file:~/src-d/spark-api/target/scala-2.11/test-classes/siva-files/05893125684f2d3943cd84a7ab2b75e53668fba1.siva to /tmp/siva-files/05893125684f2d3943cd84a7ab2b75e53668fba1.siva
   Copy file:~/src-d/spark-api/target/scala-2.11/test-classes/siva-files/fff7062de8474d10a67d417ccea87ba6f58ca81d.siva to /tmp/siva-files/fff7062de8474d10a67d417ccea87ba6f58ca81d.siva
   Copy file:~/src-d/spark-api/target/scala-2.11/test-classes/siva-files/fff7062de8474d10a67d417ccea87ba6f58ca81d.siva to /tmp/siva-files/fff7062de8474d10a67d417ccea87ba6f58ca81d.siva
   Copy file:~/src-d/spark-api/target/scala-2.11/test-classes/siva-files/fff840f8784ef162dc83a1465fc5763d890b68ba.siva to /tmp/siva-files/fff840f8784ef162dc83a1465fc5763d890b68ba.siva
   Copy file:~/src-d/spark-api/target/scala-2.11/test-classes/siva-files/fff840f8784ef162dc83a1465fc5763d890b68ba.siva to /tmp/siva-files/fff840f8784ef162dc83a1465fc5763d890b68ba.siva
   ```
   What seems to be a sign of excessive copy. Interesting enough, this does not happen with another test in the same spec.

[...]

PR #28, opened 2017-09-13
Add atomic get method to avoid race conditions

ajnavarro:


bzz:
This should solve #26 

is there anything left, before merging this guy?

PR #38, opened 2017-09-18
Add preliminary Dockerfile

ajnavarro:
All notebooks inside examples folder will be included on the image.

To run the container, the preliminary command is: docker run -p 8888:8888 -v /path/to/my/root-repositories:/repositories [CONTAINER_NAME]


bzz:
@ajnavarro do you think we can publish it in `quay.io/srcd`, so it's becomes reproducible?

bzz:
Will merge if there is no objections - as title says it's (awesome) preliminary image, and all further automation could be handled in subsequent PRs

PR #41, opened 2017-09-18
UAST: first impl - UDF using bblfsh/clinet-scala

bzz:
This is initial approach of implementing `.parseUASTs()` using a custom UDF delegating to [bblfsh/client-scala](https://github.com/bblfsh/client-scala)

This implementation is sub-optimal, as it does not re-use bblfsh client instances between rows. Following a better approach from scratch is bit tricky: it consists of `.mapPartitions` over the DF with some schema modification (adding a column), and requires more time.
 
This approach is nevertheless useful, as both implementation exhibit the same API and thus can be wrapped in pySpark asap, and optimized independently.

Includes
 - [x] bblfsh server to CI
 - [x] `.parseUAST()` \w UDF
 - [x] ability to do `./sbt dependencyTree` (required to resolve dependencies conflicts)
 - [x] small bugfix for `Implicits.checkCols`

Missing things:
 - [x] tests
 - [x] parametrize bblfsh server address/port AKA pass non-column arguments to UDF
 - [x] switch to original bblfsh/client-scala, as soon as https://github.com/bblfsh/client-scala/pull/1 is merged
 - [x] ~performance optimization: re-use connections~ Will be explored in sub-sequent PRs.


bzz:
Rebased on latest master

bzz:
Thank you for the prompt reviews!

bzz:
It's ready to be merged.

src/main/scala/tech/sourced/api/Implicits.scala
```diff
@@ -1,14 +1,16 @@
...
-import org.apache.spark.sql.functions._
 import org.apache.spark.sql.{DataFrame, SparkSession}
-import tech.sourced.api.customudf.{CustomUDF, ClassifyLanguagesUDF}
+import tech.sourced.api.udf.{ExtractUASTsUDF, ClassifyLanguagesUDF, CustomUDF}
 
 object Implicits {
 
   implicit class SessionFunctions(session: SparkSession) {
     def registerUDFs(): Unit = {
+      ExtractUASTsUDF.bblfshHost = session.sparkContext.getConf.get("tech.sourced.bblfsh.grpc.host", "0.0.0.0")
```
erizocosmico: I'm gonna add functions to set this in `SparkAPI` in #45, maybe we should wait until it's merged to reuse the key constants

erizocosmico:
@bzz Just stopped the build because it has been stuck for almost 30m

bzz:
Rebased on latest master. But CI takes forever :/

bzz:
Rebased on #48 which should be merged first

bzz:
All done now, CI is green, merging if there is no further discussion

PR #42, opened 2017-09-19
Added language detection

mcarmonaa:
Currently is needed to have [enry-java](https://github.com/src-d/enry/tree/master/java) in your local ivi2 repo.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,8 +15,11 @@ lazy val root = (project in file(".")).
     libraryDependencies += jgit % Compile,
     libraryDependencies += siva % Compile,
     libraryDependencies += commonsIO % Compile,
+    libraryDependencies += "tech.sourced" % "enry-java" % "1.0",
 
     resolvers += "jitpack" at "https://jitpack.io",
+    // TODO: remove this local resolver when enry-java will be available from jitpack.
+    resolvers += "Local Ivy repository" at "file://" + Path.userHome.absolutePath + "/.ivy2/repository",
 
     test in assembly := {}
   )
--- a/src/main/scala/tech/sourced/api/Implicits.scala
+++ b/src/main/scala/tech/sourced/api/Implicits.scala
@@ -1,12 +1,24 @@
 package tech.sourced.api
 
 import org.apache.spark.SparkException
-import org.apache.spark.sql.{DataFrame, SparkSession}
 import org.apache.spark.sql.functions._
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import tech.sourced.api.customudf.{CustomUDF, ClassifyLanguagesUDF}
 
 object Implicits {
-  implicit def addSessionFunctions(session: SparkSession) = {
-    new SessionFunctions(session)
+
+  implicit class SessionFunctions(session: SparkSession) {
+    def registerUDFs(): Unit = {
+      SessionFunctions.UDFtoRegister.foreach(customUDF => session.udf.register(customUDF.name, customUDF.function))
+    }
+
+    def getRepositories(): DataFrame = Implicits.getDataSource("repositories", session)
+  }
+
+  object SessionFunctions {
+    val UDFtoRegister = List[CustomUDF](
+      ClassifyLanguagesUDF
+    )
   }
 
   implicit class ApiDataFrame(df: DataFrame) {
@@ -39,6 +51,11 @@ object Implicits {
         filesDf
       }
     }
+
+    def classifyLanguages: DataFrame = {
+      Implicits.checkCols(df, "is_binary", "path", "content")
+      df.withColumn("lang", ClassifyLanguagesUDF.function('is_binary, 'path, 'content))
+    }
   }
 
   def getDataSource(table: String, session: SparkSession): DataFrame =
@@ -51,8 +68,6 @@ object Implicits {
       throw new SparkException("method cannot be applied into this DataFrame")
     }
   }
+
 }
 
-class SessionFunctions(session: SparkSession) {
-  def getRepositories: DataFrame = Implicits.getDataSource("repositories", session)
-}
--- a/src/main/scala/tech/sourced/api/customudf/CustomUDF.scala
+++ b/src/main/scala/tech/sourced/api/customudf/CustomUDF.scala
@@ -1,0 +1,23 @@
+package tech.sourced.api.customudf
+
+import org.apache.spark.sql.expressions.UserDefinedFunction
+import org.apache.spark.sql.functions.udf
+import tech.sourced.enry.Enry
+
+
+abstract class CustomUDF {
+  val name: String
+  val function: UserDefinedFunction
+}
+
+object ClassifyLanguagesUDF extends CustomUDF {
+  val name = "classifyLanguages"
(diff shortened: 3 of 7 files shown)


bzz:
Awesome 🎉 
Let me try it out locally and get back to you later today.

CI fails on missing enry-java dependency.

We can not use jitpack for enry-java as the default build would most probably not be a fatjar one, with all the dependencies. Until https://github.com/src-d/enry/issues/102 is resolved, you could add an existing helper script [./local-install-enry-java.sh](https://github.com/bzz/berserker/blob/14ca20c75d3134d641411d7537ce230f14f981b6/local-install-enry-java.sh) to `.travis.yml` as a workaound.

src/test/scala/tech/sourced/api/BaseSparkSpec.scala
```diff
@@ -14,6 +15,12 @@ trait BaseSparkSpec extends BeforeAndAfterAll {
       .appName("test").master("local[*]")
       .config("spark.driver.host", "localhost")
       .getOrCreate()
+
+    Logger.getLogger("org").setLevel(Level.OFF)
+    Logger.getLogger("akka").setLevel(Level.OFF)
```
bzz: Do you think it's still useful? Right now, logging for tests is muted and configured though [log4j.properties](https://github.com/src-d/spark-api/blob/master/src/test/resources/log4j.properties)

src/test/scala/tech/sourced/api/BaseSparkSpec.scala
```diff
@@ -14,6 +15,12 @@ trait BaseSparkSpec extends BeforeAndAfterAll {
...
+
+    import Implicits._
+    ss.registerUDFs()
```
bzz: Not sure, but is there a reason we want to expose this to clients? 
Are there any downsides in this, being always done by `Implicits`, so user does not need to do that ?

erizocosmico:
Maybe we can get enry via jitpack for now. It "should" work after this, I think https://github.com/src-d/enry/pull/98

bzz:
👍 @erizocosmico I guess it also depends, if jitpack uses `sbt assembly` or not. Definitely worth a try!

erizocosmico:
@bzz ah, right, then probably it doesn't work. We might need to look into making the assembly build with the normal packaging task

src/test/scala/tech/sourced/api/BaseSparkSpec.scala
mcarmonaa: Since the spark-session is an object outside our library, and `Implicits` allow us augment functionality of some classes performing conversions from those classes to our own defined classes, there's no way to register automatically the udfs just importing `Implicits`. So is needed to do it by hand before use them.

src/test/scala/tech/sourced/api/BaseSparkSpec.scala
mcarmonaa: Indeed it doesn't seem so much useful now, I'll remove these lines.

mcarmonaa:
I performed changes to follow the design document, so now the exported method is `classifyLanguages`, and I removed the logger lines too.

src/main/scala/tech/sourced/api/Implicits.scala
```diff
@@ -39,6 +51,11 @@ object Implicits {
         filesDf
       }
     }
+
+    def classifyLanguages: DataFrame = {
+      Implicits.checkCols(df, "file_hash")
+      df.withColumn("lang", ClassifyLanguagesUDF.function(col("is_binary"), col("path"), col("content")))
```
bzz: Is there a reason why `file_hash` column is checked to be present in df, rather then `is_binary`, `path` and `content` that are actually used below?

src/main/scala/tech/sourced/api/customudf/CustomUDF.scala
```diff
@@ -0,0 +1,30 @@
...
+  val function: UserDefinedFunction
+}
+
+object ClassifyLanguagesUDF extends CustomUDF {
+  val name = "classifyLanguages"
+  val function = udf[Option[String], Boolean, String, Array[Byte]](getLanguage)
+
+  private def getLanguage(isBinary: Boolean, path: String, content: Array[Byte]): Option[String] = isBinary match {
+    case false => {
+      val lang = Enry.getLanguage(path, content)
```
bzz: Probably a nitpick, but do you think 

```scala
... = isBinay match {
  case true => // ...
  case false => None
}
```

has advantages over

```scala
... = if (isBinary) {
  // ...
 } else {
  None
 }
```
?

bzz:
@mcarmonaa nice job! Few more comments above.

BTW, do you see any reason not to make CI pass with .sh script?

src/main/scala/tech/sourced/api/Implicits.scala
mcarmonaa: Oops!, fixing in progress

src/main/scala/tech/sourced/api/customudf/CustomUDF.scala
mcarmonaa: No I don't, I just wanted to use new scala patterns I learnt, but probably is more readable use an `if` expression here. I'll change it

bzz:
Regarding
> ss.registerUDFs()

Do you think this approach https://github.com/src-d/spark-api/pull/41/commits/f31b1e53d7bb30c78a41f8ec5e6c149d6fae5131 would not work for some reason? 

I think it may be reasonable to have a convention of UDFs registering themselves, thus removing this "cognitive burden" from the API users.

mcarmonaa:
@bzz about https://github.com/src-d/spark-api/commit/f31b1e53d7bb30c78a41f8ec5e6c149d6fae5131, Is it ok register the udf in the catalog each time you use it?

mcarmonaa:
As talking irl with @bzz, by the moment
> ss.registerUDFs()
still be there.

The rest of the feedback has been addressed

bzz:
@mcarmonaa awesome work! Let's wait for CI to become green

bzz:
and 👍 for having nice tests!

build.sbt
```diff
@@ -15,8 +15,10 @@ lazy val root = (project in file(".")).
     libraryDependencies += jgit % Compile,
     libraryDependencies += siva % Compile,
     libraryDependencies += commonsIO % Compile,
+    libraryDependencies += "tech.sourced" % "enry-java" % "1.0",
 
     resolvers += "jitpack" at "https://jitpack.io",
+    resolvers += "Local Ivy repository" at "file://"+Path.userHome.absolutePath+"/.ivy2/repository",
```
erizocosmico: I would add a TODO or FIXME to explain why this is here, just in case we forget to delete it after enry is published

src/main/scala/tech/sourced/api/customudf/CustomUDF.scala
```diff
@@ -0,0 +1,29 @@
...
+      if (lang != "") Some(lang) else None
+    }
+}
+
+
+
+
+
+
```
erizocosmico: leftover?

mcarmonaa: Yes it is, I'll remove it

PR #43, opened 2017-09-19
[DOC] Installation and How to Run spark-api in README

mcarmonaa:
This PR is related to https://github.com/src-d/spark-api/issues/34.

A first approach to a explanation on how to install and run spark-api, pyspark examples should be added too. Feedback is welcome!

PR #45, opened 2017-09-19
implicits: make SparkAPI the entrypoint instead of using an implicit

erizocosmico:
I think adding this function to the spark session simplifies a bit the experience of the user of the library.

Before

```scala
spark.sqlContext.setConf("tech.sourced.api.repositories.path", "/path/to/files")
```

Now

```scala
spark.setRepositoriesPath("/path/to/files")
```

We could even check if the path exists and contains siva files there and so on in the future.

### DIFF
--- a/src/main/scala/tech/sourced/api/Implicits.scala
+++ b/src/main/scala/tech/sourced/api/Implicits.scala
@@ -7,12 +7,19 @@ import tech.sourced.api.customudf.{CustomUDF, ClassifyLanguagesUDF}
 
 object Implicits {
 
+  val repositoriesPathKey = "tech.sourced.api.repositories.path"
+
   implicit class SessionFunctions(session: SparkSession) {
     def registerUDFs(): Unit = {
       SessionFunctions.UDFtoRegister.foreach(customUDF => session.udf.register(customUDF.name, customUDF.function))
     }
 
     def getRepositories(): DataFrame = Implicits.getDataSource("repositories", session)
+
+    def setRepositoriesPath(path: String): SparkSession = {
+      session.sqlContext.setConf(Implicits.repositoriesPathKey, path)
+      session
+    }
   }
 
   object SessionFunctions {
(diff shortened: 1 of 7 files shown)


bzz:
Nice! 

And how about pushing it a step further and making Python and Scala experience of using SparkAPI looking more similar?

In Scala this could mean adding a single entrypoint object that could hide some useful imports/configurations/etc, something like what was proposed in https://github.com/src-d/spark-api/pull/43#issuecomment-330501672:

```scala
val api = SparkAPI(spark, "path/to/siva/files")

api.getRepositories()
```

What do you think?

erizocosmico:
Sounds great!

erizocosmico:
Done @bzz!

Now it looks like this:

```scala
import tech.sourced.api.SparkAPI

val reposDf = SparkAPI(spark, "/path/to/foo").getRepositories
```

bzz:
Looks awesome to me!

bzz:
As soon as #43 is merged, shall this be rebased and example in README updated to use this as well?

erizocosmico:
Yup!

One hairy thing that I noticed is that you now have to make 2 imports:

```scala
import tech.sourced.api.SparkAPI
import tech.sourced.api.Implicits._
```

Perhaps we could do something like:

```scala
package api {
  // define spark api
  // define implicits
}
```
Then it would become

```
import tech.sourced.api._
```

bzz:
Oh, yes, that's a great catch - would be rally nice to have single `import tech.sourced.api._`!

erizocosmico:
Working on that, then

erizocosmico:
Fixed!

Moved implicits to a package object and made some things package-private to not make the user import a bunch of useless internal stuff.

Anyway, if the user reeeeeally wants to specify it can be done like so:

```scala
import tech.sourced.api.{SparkAPI, ApiDataFrame}
```

erizocosmico:
Added documentation for the user-facing API as well.

bzz:
The docs looks awesome. 


>One downside of this approach is that Java-generated classes look like:

Yes, but as soon as Java API is not the focus and it works in Python - that should be fine.

Let's get it merge then?

erizocosmico:
:+1:

:shipit:

PR #46, opened 2017-09-20
Doc: update high-level description in README

bzz:
Few high-level details on current architecture of Spark API


README.md
```diff
@@ -1,8 +1,18 @@
 # spark-api [![Build Status](https://travis-ci.org/src-d/spark-api.svg?branch=master)](https://travis-ci.org/src-d/spark-api)
 
-High-level Spark API is for running scalable data retrieval pipelines that process and manipulate any number of code repositories for source code analysis. Written mostly in Scala, it aims to be robust, friendly and flexible: it is built on top of Apache Spark, accessible both via Scala and Python Spark APIs, and capable of running on large-scale distributed clusters over petabytes of data.
+**Spark API** is a library for running scalable data retrieval pipelines that process any number of Git repositories for source code analysis.
```
erizocosmico: shouldn't it be data processing instead of data retrieval? AFAIK spark-api does not fetch anything

bzz: Good point. Although right now it does not, eventually it will have an option to clone the repositories.

Also 'retrieval' traditionally stand for - "obtaining structured representation" in IR discipline..which is also applicable in our case, so I think it may be fine. 

But I do not have strong opinion, will be happy to change if you still think "processing" it's better!

erizocosmico: It actually makes sense, let's keep it this way then

PR #39, opened 2017-09-18
python: SparkAPI python wrapper

erizocosmico:
First draft of the SparkAPI python wrapper.

### What's missing?

* [x] Tests! Although it has been tested IRL, it might be good to add some tests. But when the API is stable (or as stable as it can be) they can be added. After all, it's a very thin layer.
* [x] Auto-add and manage the SparkAPI jar. **Can't be done**
*  README. Should this package have it's own README or should all its contents be on the main README? Being done in #43 
* [x] travis jobs
* [x] pypi release on tag with travis
* [x] Update python notebook Dockerfile demo **(blocked until #38 is merged)**

### Example

```python
from sourced.spark import API as SparkAPI
from pyspark.sql import SparkSession

spark = SparkSession.builder.appName("test").master("local[*]").getOrCreate()
api = SparkAPI(spark, '/path/to/siva/files')
api.repositories.filter("url = 'yadayada.foo/url.git'").references.head_ref.commits.show()
spark.stop()
```

* Install the python wrapper doing `python setup.py install` or `python setup.py install --user` or add the `python` folder of the `spark-api` repo to your `PYTHONPATH` (`export PYTHONPATH="$PYTHONPATH:/path/to/python/folder"`).
* Run with `spark-submit --jars path/to/spark-api-jar /path/to/this/example.py`

  Can also be run from pyspark shell (adding the `--jar /path/to/spark-api-jar` when starting the shell). In the case of the pyspark shell you don't need to create the spark session, it's already available as `spark`.

### Design decisions

Because with Py4J and Python we lose the Scala implicits we lose the ability to automatically add our methods to `DataFrame`. Because of that, we have a `SourcedDataFrame` which is a subclass of `pyspark.sql.DataFrame`. But because the methods return new `DataFrame`s whenever we use a method of `DataFrame` we are stuck with a plain `DataFrame`.

For example, doing the following would not be valid:

```python
api.repositories.filter("some condition").references
```

Because `DataFrame` does not have a `references` method or property.

To overcome this, `SourcedDataFrame` overrides the methods of `DataFrame` that return a new `DataFrame` and transforms them to the corresponding class. For example, if we have an instance of `RepositoriesDataFrame`, any method that returns a new `DataFrame` now will return an instance of `RepositoriesDataFrame` instead, and so on.

[...]

PR #48, opened 2017-09-21
Fix: race condition on copy

bzz:
Changes:
 - adds synchronization around the repository copy
 - adds logging to RepositoryProvider

Motivation:
#28 with atomic `Map.getOrElseUpdate` might be not enough, as it still can run file copy multiple times.

According to the [`MapLike.html#getOrElseUpdate`](https://www.scala-lang.org/api/current/scala/collection/mutable/MapLike.html#getOrElseUpdate(key:K,op:=>V):V)  docs:
> Concurrent map implementations may evaluate the expression op multiple times, or may evaluate op without inserting the result.

### DIFF
--- a/src/main/scala/tech/sourced/api/provider/RepositoryProvider.scala
+++ b/src/main/scala/tech/sourced/api/provider/RepositoryProvider.scala
@@ -8,6 +8,7 @@ import org.apache.commons.io.FileUtils
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.spark.input.PortableDataStream
+import org.apache.spark.internal.Logging
 import org.eclipse.jgit.lib.{Repository, RepositoryBuilder}
 import tech.sourced.api.util.MD5Gen
 import tech.sourced.siva.SivaReader
@@ -15,12 +16,12 @@ import tech.sourced.siva.SivaReader
 import scala.collection.JavaConverters._
 import scala.collection.concurrent
 
-class RepositoryProvider(val localPath: String) {
+class RepositoryProvider(val localPath: String) extends Logging {
   private val repositories: concurrent.Map[String, Repository] =
     new ConcurrentHashMap[String, Repository]().asScala
 
   def get(conf: Configuration, path: String): Repository =
-    repositories.getOrElseUpdate(path, RepositoryProvider.genRepository(conf, path, localPath))
+    repositories.getOrElseUpdate(path, genRepository(conf, path, localPath))
 
   def get(pds: PortableDataStream): Repository =
     this.get(pds.getConfiguration, pds.getPath())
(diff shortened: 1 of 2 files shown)


src/main/scala/tech/sourced/api/provider/RepositoryProvider.scala
```diff
@@ -8,19 +8,21 @@ import org.apache.commons.io.FileUtils
...
 
-class RepositoryProvider(val localPath: String) {
+class RepositoryProvider(val localPath: String) extends Logging {
   private val repositories: concurrent.Map[String, Repository] =
     new ConcurrentHashMap[String, Repository]().asScala
 
-  def get(conf: Configuration, path: String): Repository =
-    repositories.getOrElseUpdate(path, RepositoryProvider.genRepository(conf, path, localPath))
+  def get(conf: Configuration, path: String): Repository = synchronized {
+    repositories.getOrElseUpdate(path, genRepository(conf, path, localPath))
```
mcarmonaa: Is not the problem that `getOrElseUpdate` could evaluate `genRepository` although it's not necessary to insert it? Does `synchronized` fix that?

bzz: Good question @mcarmonaa  - after re-reading the docs literally, I see why you have it.

The problem is that `genRepository()` has side-effects, and we want it to be triggered only once per task (thread).

Let's think, in what circumstances the concurrent Map implementation of `getOrElseUpdate` might behave as advertised (evaluate `genRepository ` more then once)?

It most probably going to happen in case of concurrent access to this method by multiple threads. `synchronized`  does not fix anything per-se but would rather prevent such access pattern to the Map.

Let me know if that answers your question, or if you have ideas on the better way of changing this behaviour!

mcarmonaa: I see, but I was thinking about what happens if the same thread calls `get` several times and `genRepository` is evaluated several times too when the key is already in the map since:

> ... or may evaluate op without inserting the result.

bzz: I see!

But looking at the code, I can not see how this can be possible for a single thread to do so, esp. in case when the key is already present..

```scala
  def getOrElseUpdate(key: A, op: => B): B =
    get(key) match {
      case Some(v) => v
      case None => val d = op; this(key) = d; d
    }
```

erizocosmico: If `genRepository` is the issue here, why not synchronize that method in particular?

Also, a fix for what Manuel says could be checking first if it exists and update only if it doesn't. That way we avoid evaluating op even if it's not to be inserted.

Moreover, the docs say:

> Concurrent map implementations **may evaluate the expression `op` multiple times**

So even if it should evaluate once it may evaluate it more than once.

Something like this should avoid all that:

```scala
  def get(conf: Configuration, path: String): Repository = synchronized {
    repositories.get(path) match {
      case Some(repo) => repo
      case None => {
        val repo = RepositoryProvider.genRepository(conf, path, localPath)
        repositories.put(path, repo)
        repo
      }
    }
  }
```

bzz: And that was the original implementation :)
let me switch it back

bzz:
Done. \cc @erizocosmico @mcarmonaa for another round

bzz:
CI on #39 failed again with this https://travis-ci.org/src-d/spark-api/jobs/278127003


```
[info]   Cause: java.io.FileNotFoundException: /tmp/siva-files/05893125684f2d3943cd84a7ab2b75e53668fba1.siva (No such file or directory)
```

bzz:
Merging if there is no further discussion

PR #55, opened 2017-09-22
iterator: implement CleanupIterator to clean unpacked resources

### LINKED ISSUE #29: Cleanup un-packed .siva files
After [copying from HDFS and unpacking a .siva files](https://github.com/src-d/spark-api/blob/master/src/main/scala/tech/sourced/api/provider/RepositoryProvider.scala#L80) at the start of the job, local FS on workers has to be cleaned-up at when the Job ends.

Current implementation does not take care of it.

There are two ways to do that
 - `CompletionIterator` / `InterruptibleIterator` preferable, see [example](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCRDD.scala#L304)
 - TaskContext listener

This issue is about picking a simplest one that would feat Spark API use cases and implementing it.
Unpacked .siva file optimization needs to be taken into account.

erizocosmico:
Fixes #29 

### Design decisions

I implemented a `CleanupIterator`, which is basically an iterator extending `InterruptibleIterator` that executes some cleanup code when there is an exception (`Throwable` because `TaskKilledException` is not public :disappointed: )

Whenever a repository is fully iterated and done with (or something was thrown from inside), it's closed and its unpacked files deleted if there is nobody else using it. Otherwise, just keeps them open until the last one using them closes the repo.

I don't personally fancy much the refcount approach for this, but you still need it if you add a listener on the task context, because multiple tasks can be using the same repo. The iterator approach I felt it was cleaner than the task context one and `CompletionIterator` is private. Anyway, `CleanupIterator` is basically a `CompletionIterator` that cleans up when there is an exception as well.Fixes #29

### DIFF
--- a/src/main/scala/tech/sourced/api/DefaultSource.scala
+++ b/src/main/scala/tech/sourced/api/DefaultSource.scala
@@ -1,11 +1,11 @@
 package tech.sourced.api
 
-import org.apache.spark.SparkException
+import org.apache.spark.{InterruptibleIterator, SparkException, TaskContext}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.{Row, SQLContext}
-import tech.sourced.api.iterator.{BlobIterator, CommitIterator, ReferenceIterator, RepositoryIterator}
+import tech.sourced.api.iterator._
 import tech.sourced.api.provider.{RepositoryProvider, SivaRDDProvider}
 import tech.sourced.api.util.ColumnFilter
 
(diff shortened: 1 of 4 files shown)


erizocosmico:
@bzz you mean split the setting in two to control each one of them? Another option could be not copying the siva file if there are unpacked files already.

ajnavarro:
>  you mean split the setting in two to control each one of them? Another option could be not copying the siva file if there are unpacked files already.

Sounds good.

erizocosmico:
@ajnavarro which one of them?

ajnavarro:
@erizocosmico  Not copying the siva file if we already unpacked them.

erizocosmico:
Done @ajnavarro

erizocosmico:
Should we merge? @ajnavarro

PR #69, opened 2017-09-29
return all commit files in BlobIterator, implement getFiles in api

### LINKED ISSUE #62: [API] Create a new method to get files directly from repositories df
We need a new method that would get files from repositories using several filters to improve performance.

The first step is to add two new columns to files relation:
- repository_id
- reference_name

Then, we should create a new api method, something like:

```scala 
def getFiles(repositoriesIds:Seq[String],referenceNames:Seq[String],commitHashes:Seq[String]):DataFrame
```

### LINKED ISSUE #68: BlobIterator does not return the files of all commits
Right now, BlobIterator only returns the files of the commits pointed to by a reference, instead of all commits in a reference.

erizocosmico:
Closes #68 
Closes #62 

BlobIterator now returns the files of each commit of each ref of each repo, instead of just the files on all ref head.

The schema of the files table is also changed to contain repository_id and reference_name so the filters are pushed down to the iterator and some optimizations can be done for the retrieval of just a small amount of files with certain restrictions such
as just retrieving files of a few hashes or references.
That said, this is just a temporary hack until more clever measures are put in place, because now we have two columns that will always be null and will always be dropped from the dataframe just for the sake of getting the filters pushed down.

SparkAPI has now a method `getFiles` to retrieve files given a set of repositories, references and/or hashes which takes advantage of the aforementioned optimizations until a more general approach is implemented by just using the regular dataframe of files.

### Note

This is still a work in progress. A few methods still need to be implemented and lacks a bit of testing, but I sent the PR to start getting feedback specially on the `BlobIterator` side.

CI does not pass yet.

### DIFF
--- a/src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
+++ b/src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
@@ -22,18 +22,12 @@ class BlobIterator(requiredColumns: Array[String], repo: Repository, filters: Ar
   extends RootedRepoIterator[CommitTree](requiredColumns, repo) with Logging {
 
   override protected def loadIterator(): Iterator[CommitTree] = {
-    val repositories = mutable.Buffer[String]()
-    val references = mutable.Buffer[String]()
-    val hashes = mutable.Buffer[String]()
+    val f = mutable.HashMap[String, mutable.Buffer[String]]()
 
     filters.flatMap(_.matchingCases)
       .foreach(filter => filter match {
-        case ("repository_id", repos) =>
-          repositories ++= repos.map(_.asInstanceOf[String])
-        case ("reference_name", refs) =>
-          references ++= refs.map(_.asInstanceOf[String])
-        case ("commit_hash", commitHashes) =>
-          hashes ++= commitHashes.map(_.asInstanceOf[String])
+        case (k@("repository_id" | "reference_name" | "commit_hash"), values) =>
+          f(k) = mutable.Buffer(values.map(_.asInstanceOf[String]): _*)
         case anyOtherFilter =>
           log.debug(s"BlobIterator does not support filter $anyOtherFilter")
       })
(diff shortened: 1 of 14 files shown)


src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -25,51 +22,86 @@ class BlobIterator(requiredColumns: Array[String], repo: Repository, filters: Ar
...
+      filesFromHashes(hashes)
     } else {
-      log.debug(s"Iterating all ${refs.size} refs")
-      refs.toIterator.flatMap { ref =>
-        log.debug(s" $ref")
-        JGitBlobIterator(getTreeWalk(ref.getObjectId), log)
+      var repos = repo.getConfig.getSubsections("remote").asScala
+        .map(getRepositoryId(_).get)
+
+      if (!repositories.isEmpty) {
```
ajnavarro: .nonEmpty

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -25,51 +22,86 @@ class BlobIterator(requiredColumns: Array[String], repo: Repository, filters: Ar
...
+
+      var refs = repo.getAllRefs.asScala.values.toIterator
+        .filter(ref => {
+          val (repoId, refName) = parseRef(ref.getName)
+          repos.contains(repoId)
+        })
+
+      // If there are references to filter, just get the hash of the commit they point to
+      // and treat it as a hash filter.
+      if (!references.isEmpty) {
```
ajnavarro: .nonEmpty

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -25,51 +22,86 @@ class BlobIterator(requiredColumns: Array[String], repo: Repository, filters: Ar
...
+          contained
+        })
+      }
+
+      if (refs.isEmpty) {
+        Seq().toIterator
+      } else {
+        hashes.distinct
+        var commits = CommitIterator.refCommits(repo, refs.toList.distinct: _*)
+        if (!hashes.isEmpty) {
```
ajnavarro: .nonEmpty

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -25,51 +22,86 @@ class BlobIterator(requiredColumns: Array[String], repo: Repository, filters: Ar
...
-              JGitBlobIterator(getTreeWalk(commitId), this.log)
-            } else {
-              Seq()
-            }
-          }
+    val repositories = mutable.Buffer[String]()
+    val references = mutable.Buffer[String]()
+    val hashes = mutable.Buffer[String]()
+
+    filters.flatMap(_.matchingCases)
```
ajnavarro: Instead of fill 3 mutable Buffers, we can create a map with "repository_id", "reference_name" and "commit_hash" as keys, and buffers as values.

src/main/scala/tech/sourced/api/iterator/BlobIterator.scala
```diff
@@ -25,51 +22,86 @@ class BlobIterator(requiredColumns: Array[String], repo: Repository, filters: Ar
...
-          Seq()
-      }
+      })
 
-    if (filtered.hasNext) {
-      filtered
+    // Fast path, if there are no repositories or references to filter by
+    // just return the files from the given hashes, which is the fastest
+    // option of all.
+    if (!hashes.isEmpty && repositories.isEmpty && references.isEmpty) {
```
ajnavarro: Having a map, you can check if the map contains the needed keys to continue for this if branch.

erizocosmico:
Done @ajnavarro 

One thing that might be problematic about this approach is the following:

Imagine we have:
* a branch HEAD with commits A -> B -> C -> D -> E
* a branch develop with (HEAD) -> F -> G
* a branch foo with (HEAD) -> H
* Each commit contains only 1 file

We'll have 18 rows:

A, B, C, D, E x 3 because that commit is in each reference, F and G, which are in develop and H, which is in foo.

This is because `reference_name` and `repository_id` need to be filled if we want to filter by them, even if we drop them afterwards. The """solution""" to this is the `.distinct` in `getFiles`, but at that point all of the files have been read repeatedly.
Previous implementation only loaded a commit once, but that requires to not fill `reference_name` and `repository_id`, which makes it impossible to filter by those fields.

erizocosmico:
`getFiles` has been implemented so that no filter is required. There is quite a few ways of calling it:

- `api.getFiles`: same as doing `api.getRepositories.getReferences.getCommits.getFiles`
- `api.getFiles(repositoryIds = ids)` returns files of only these repositories (fast)
- `api.getFiles(referenceNames = names)` returns files of only those refs (slow)
- `api.getFiles(commitHashes = hashes)` returns files of only those hashes (slow)
- `api.getFiles(repositoryIds = ids, referenceNames = refs)` returns files in references in given repos (fast)
- `api.getFiles(repositoryIds = ids, commitHashes = hashes)` returns files of hashes in given repos (fast)
- `api.getFiles(referenceNames = refs, commitHashes = hashes)` returns files of hashes in given references (slow)
- `api.getFiles(repositoryIds = ids, referenceNames = refs, commitHashes = hashes)` returns files of hashes in references in given repos (fastest)

TL;DR: if you provide `repositoryIds` it's fast, otherwise it's slow because it has to iterate over everything in all repos.

With this API we add only a single method with a lot of different possible usages.

erizocosmico:
Done @ajnavarro @mcarmonaa 

Also adds documentation to all the python api.

PR #70, opened 2017-10-02
[WIP] Add a new rule to avoid unnecessary Joins

ajnavarro:
Things to do
- Implement unhandledFilters method to just push up the neccessary filters to spark, by default all filters are executed on spark side again. Doing this we avoid too much data on requiredColumns relation value.
- Change ColumnFilter to work with Expressions instead of Filters, because we changed from PrunedFilteredScan relation to CatalystScan relation.
- Add new logic to GitRelation to concatenate iterators depending of the conditions and filters.
- Add a lot of tests to check several use cases.
- Add metadata to attributes to know the source of them (repositories,references,commits,files).

### DIFF
--- a/src/main/scala/tech/sourced/api/DefaultSource.scala
+++ b/src/main/scala/tech/sourced/api/DefaultSource.scala
@@ -1,24 +1,30 @@
 package tech.sourced.api
 
-import org.apache.spark.{InterruptibleIterator, SparkException, TaskContext}
+import org.apache.spark.SparkException
 import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.catalyst.expressions.{Attribute, Expression}
 import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.{Row, SQLContext}
 import tech.sourced.api.iterator._
 import tech.sourced.api.provider.{RepositoryProvider, SivaRDDProvider}
-import tech.sourced.api.util.ColumnFilter
 
 class DefaultSource extends RelationProvider with DataSourceRegister {
 
-  override def shortName() = "git"
+  override def shortName = "git"
 
   override def createRelation(sqlContext: SQLContext, parameters: Map[String, String]): BaseRelation = {
     val table = parameters.getOrElse(DefaultSource.tableNameKey, throw new SparkException("parameter 'table' must be provided"))
-    val path = parameters.getOrElse(DefaultSource.pathKey, throw new SparkException("parameter 'path' must be provided"))
-    val localPath = sqlContext.getConf("spark.local.dir", "/tmp")
 
-    GitRelation(sqlContext, table, path, localPath)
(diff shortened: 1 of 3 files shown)


src/main/scala/tech/sourced/api/GitOptimizer.scala
```diff
@@ -0,0 +1,135 @@
...
+import org.apache.spark.sql.catalyst.plans.{Inner, JoinType}
+import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.execution.datasources.LogicalRelation
+import org.apache.spark.sql.types.{BooleanType, StructField, StructType}
+
+class SquashGitRelationJoin extends Rule[LogicalPlan] {
+  def apply(plan: LogicalPlan): LogicalPlan = plan transformUp {
+    // Joins are only applicable per repository, so we can push down completely the Join into the datasource
+    case q@Join(_, _, _, _) =>
+      val jd = GitOptimizer.canBeOptimized(q)
```
erizocosmico: `canBeOptimized` suggests a boolean result, but this returns something else. What do you think about naming it something else like `getOptimizableJoinData` or whatever?

src/main/scala/tech/sourced/api/DefaultSource.scala
```diff
@@ -27,40 +31,61 @@ object DefaultSource {
   val pathKey = "path"
 }
 
-case class GitRelation(sqlContext: SQLContext, tableName: String, path: String, localPath: String) extends BaseRelation with PrunedFilteredScan {
+case class GitRelation(sqlContext: SQLContext, schema: StructType, joinConditions: Option[Expression] = None) extends BaseRelation with CatalystScan {
+  private val localPath: String = sqlContext.getConf(localPathKey, "/tmp")
```
erizocosmico: what is this for?

src/main/scala/tech/sourced/api/GitOptimizer.scala
```diff
@@ -0,0 +1,135 @@
...
+        JoinData(None, valid = true, joinCondition = joinCondition, attributes = out, sqlContext = Some(sqlc))
+      case other =>
+        logWarning(s"Join cannot be optimized. Invalid node: $other")
+        JoinData()
+    }
+
+    // Reduce all filter expressions into one
+    jd.reduce((jd1, jd2) => {
+      // get all filter expressions
+      val exprOpt = (jd1.filterExpression, jd2.filterExpression) match {
```
erizocosmico: this can be a function so it can be reused with the join conditions, what do you think?

src/main/scala/tech/sourced/api/DefaultSource.scala
ajnavarro: This will be needed when all the iterator stuff will be implemented.

ajnavarro:
Obsolete. See: https://github.com/src-d/spark-api/pull/88

PR #76, opened 2017-10-03
tests: setup integration tests

erizocosmico:
Dunno how useful will be running tests from SQL queries (specially after the iterators refactor @ajnavarro is doing) but... here it is.


src/integration/scala/tech/sourced/api/IntegrationTest.scala
```diff
@@ -0,0 +1,45 @@
...
+  var ss: SparkSession = _
+  val resourcesPath = getClass.getResource("/siva-files").toString
+
+  "SQL queries" should {
+    Source.fromInputStream(getClass.getResourceAsStream("/queries.sql"))
+      .getLines
+      .foreach(sql => s"query: $sql" in {
+        val df = ss.sql(sql)
+        println(sql)
+        df.show
```
ajnavarro: we should check the output result too

erizocosmico: Then we'd need to move away from a plain SQL file and use something more complex (or add a scala test file for it)

erizocosmico:
I changed the SQL file with a `Map` to also be able to check the results. I don't know how maintainable will that be in the future if we start adding a lot of queries.

Also, I tried running these tests with other bigger siva files and it was awfully slow, so I had to resort to use the same files used right now in regular unit tests until we have the iterator fix or something.

erizocosmico:
Shall we merge or close this?

erizocosmico:
ping @ajnavarro

erizocosmico:
Rebased, merging

erizocosmico:
This is hella outdated and has a previous build config so I will close this and will put up a new PR with this once we need it with updated build conf and so on.

PR #88, opened 2017-10-09
optimize joins and make iterators chainable to improve performance

### LINKED ISSUE #63: Push down JOIN conditions if possible to the GitDatasource
Actually, if we have a query like this one:
```scala
api.getRepositories().where("id = 'github.com/foo/bar'").getHEAD.getFiles.extractUASTs().select("name", "path", "uast")
```
The optimized plan will be (simplified):

```
Project
+-Join[commit_hash = hash]
   :   +- GitRelation[FILES]
   +- Join[repository_id = id]
       :   +- Filter[name="refs/heads/HEAD",repository_id="github.com/foo/bar"]
       :       +- GitRelation[REFERENCES]
       +- Filter[id="github.com/foo/bar"]
           +- GitRelation[REPOSITORIES]
```
As we can see, no filter is pushed down to the `files` relation, having to push up all the files over all the revisions over all the repositories in this case.
We should implement a rule to be able to transform the optimized plan from above to:

```
Project
+- Filter[commit_hash = hash, repository_id = id, repository_id = "github.com/foo/bar"]
    +-GitRelation[FILES]
```
Then, the non resolved conditions (`repository_id = id` in this case) should be handled by the iterator.

erizocosmico:
Closes #63 

With the optimization rule to squash joins into a single GitRelation we have the opportunity to make the iterators chainable to reuse the previous step to compute the next instead of computing most of them a lot of times, which will bring a considerable speedup to the overall usage of the SparkAPI.
This commit renders `getFiles` obsolete, but it will be kept for now for compatibility reasons (or convenience, in some cases).
What this implements is a way for the RootedRepoIterator to manage two iterators at the same time: the internal iterator (that is, the thing being iterated at the moment) and an optional previous iterator. For every row of the previous iterator, the internal iterator is loaded for that row and fully iterated until there are no more items. Then the previous iterator is advanced again and so on until the previous iterator is empty.
So, for example, to iterate files efficiently, we would have a BlobIterator, whose prevIter would be a CommitIterator, whose prevIter would be a ReferenceIterator, whose prevIter would be a RepositoryIterator.
The benefit of this is that if we have 500 repositories, but only 1 matches, only 1 repository will make it to the ReferencesIterator and so on, instead of having to iterate every single thing for every repository.

Only very specific filters are supported at the iterator level:
- RepositoryIterator: by "id"
- ReferenceIterator: by "repository_id" and "name"
- CommitIterator: by "repository_id", "reference_name", "hash" and "index"
- BlobIterator: no specific iterator-level filters, spark will take care of it, since this is the last step in every computation

The rest of the filters (or any condition that is not an "IN" or an "=") will be handled by spark, which means it will be slower because they will be processed only to be discarded in the end.


src/main/scala/tech/sourced/api/DefaultSource.scala
```diff
@@ -50,62 +49,145 @@ object DefaultSource {
...
+    val sources = sc.broadcast(GitRelation.getSources(tableSource, schema))
+    val filtersBySource = sc.broadcast(GitRelation.getFiltersBySource(filters))
+
+    sivaRDD.flatMap(pds => {
+      val provider = RepositoryProvider(reposLocalPath.value, skipCleanup)
+      val repo = provider.get(pds)
+
+      // since the sources are ordered by their hierarchy, we can chain them like this
+      // using the last used iterator as input for the current one
+      var iter: Option[RootedRepoIterator[_]] = None
```
ajnavarro: In the future we can do this in a recursive, immutable way, but for now is ok

erizocosmico:
rebased and ready to be merged once our travis lord is green

PR #96, opened 2017-10-10
Improve Basic examples using enry, cache and aggregations

ajnavarro:


ajnavarro:
We should add some examples related with the UAST and the libUAST when ready.

PR #100, opened 2017-10-10
Avoid execute parent deploy on python execution

ajnavarro:


se7entyse7en:
Add @se7entyse7en and @dpordomingo as maintainers.

PR #106, opened 2017-10-11
Get local dir correctly

ajnavarro:


src/main/scala/org/apache/spark/UtilsWrapper.scala
```diff
@@ -0,0 +1,7 @@
+package org.apache.spark
+
+import org.apache.spark.util.Utils
+
+object UtilsWrapper extends {
```
mcarmonaa: What is extending this?

mcarmonaa:
Why must be this `UtilsWrapper` under `org.apache.spark` instead under our `util` package?

PR #121, opened 2017-10-19
sbt: upgrade client-scala version

erizocosmico:
New scala client version includes synchronization when installing the jar shared object to local filesystem, which before could lead to weird stuff


bzz:
Using https://github.com/bblfsh/client-scala/pull/27 sounds great

PR #137, opened 2017-10-24
Add end-to-end test for spark-shell

bzz:
This test runs actual `spark-shell --jars engine-uber.jar`

It's purpose is to catch any dependency miss-match at runtime that might happen due to transitive dependencies collision at build time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ cache:
   directories:
     - $HOME/.sbt
     - $HOME/.ivy2
+    - .spark-dist
 
 services:
   - docker
@@ -41,6 +42,19 @@ matrix:
             tags: true
 
     - jdk: openjdk8
+
+    - jdk: openjdk8
+
+      env:
+        - END_TO_END=true
+
+      before_script:
+        - make docker-bblfsh
+
+      script:
+        - make build
+        - ./_tools/getApacheSpark.sh "2.2.0" "2.7"
+        - ./spark/bin/spark-shell --jars ./target/engine-uber.jar -i src/test/resources/SparkShellScript.scala
 
     - language: python
       python: 2.7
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -42,6 +43,19 @@ matrix:
 
     - jdk: openjdk8
 
+    - jdk: openjdk8
```
mcarmonaa: Is there a reason to have this line repeated or is a bug?

_tools/getApacheSpark.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+# A helper script to get Apache Spark spark-shell on CI mahine
```
mcarmonaa: s/mahine/machine

src/test/resources/SparkShellScript.scala
```diff
@@ -0,0 +1,24 @@
+import tech.sourced.api._
```
mcarmonaa: s/tech.sourced.api._/tech.sourced.engine._

src/test/resources/SparkShellScript.scala
```diff
@@ -0,0 +1,24 @@
...
+
+
+//End-to-end test of packages spark-api.jar intended to be used with ./spark-shell -i SparkShellScript.scala
+//it's not part of src/test/scala to avoid compilation during build
+spark.version
+
+try {
+  val engine = engine(spark, "./src/test/resources/siva-files/")
```
mcarmonaa: s/engine(spark, "./src/test/resources/siva-files/")/Engine(spark, "./src/test/resources/siva-files/")

.travis.yml
bzz: It is intentional - it adds a new element to CI build matrix, that will run only end to end tests
screen shot 2017-10-26 at 3 44 55 pm

bzz:
https://github.com/src-d/engine/issues/142 might have been caught using the same end to end approach, only on pySpark side..

erizocosmico:
Shall we make this pass the tests and merge it or close it?

ajnavarro:
Closing it. We should think in a more general way to do integration tests in several spark deployment modes (Standalone, local, yarn and so on)

PR #143, opened 2017-10-27
build: publish fatjar on maven by default

erizocosmico:


build.sbt
```diff
@@ -1,29 +1,27 @@
...
+// grpc for bblfsh/client-scala needs to be newer then in Spark
+libraryDependencies += fixNettyForGrpc
+libraryDependencies += jgit % Compile
+libraryDependencies += siva % Compile
+libraryDependencies += bblfsh % Compile
+libraryDependencies += commonsIO % Compile
+libraryDependencies += enry % Compile
+
+test in assembly := {}
+assemblyJarName in assembly := s"${name.value}-${version.value}.jar"
```
mcarmonaa: What is the difference between having all these things grouped in a `lazy val root` variable and having them in this way?

bzz:
Nice! Which artefact gets published now?

AFAIU, it should now work with `--packages` as well as `--jars` without fetching any external dependencies from the Net

build.sbt
erizocosmico: Moved them to the root of the file while I was testing this and forgot to move them back. If you have only one project I don't think it does make a difference.

I can move it back if you guys want, but IMHO if feels more readable this way

erizocosmico:
@bzz now `engine-{version}.jar` gets published as it used to before, but it is a fatjar instead. Also, we have the `engine-uber.jar` as we used to for easier access with a constant name, which is just a copy of the main fatjar.

Issue #132, opened 2017-10-20
Perf: not cleaning unpacked .siva files makes things slower


bzz:
There is empirical evidence of what supposed to be optimization, may be causing a performance degradation - if cleaning unpack .siva files is skipped , it makes job run much slower (like 2x)

erizocosmico:
More detailed explanation of what happened:

- Using skip cleanup optimization the first run was way slower than without it.
- Subsequent runs took as long as without the optimization

mcarmonaa:
After `skipCleanup` option fix, I performed test in local with the [100 samples of siva files](https://drive.google.com/open?id=0BxNVgwtOUkMUaG5wbmpnRklmMEk)  and I couldn't appreciate performance differences:

```text
LOCAL
---------

- skip_cleanup=False
- Wall time:
first time it ran: 4.92s
second time it ran: 2.65s

- skip_cleanup=True
- Wall time:
first time it ran: 5.06s
second time it ran: 2.24s
```

The snippet of code from the `engine-jupyter` notebook example which was used for testing was:

```python
%%time
engine.repositories.filter("is_fork = false")\
.select("id").distinct()\
.count()
```

Trying to run the same test in the cluster fails with `java.io.IOException: No space left on device` because not removing siva files fills the disk.

ajnavarro:
Can we close this issue then?

PR #144, opened 2017-10-27
Fixed skipCleanup option

mcarmonaa:
***This PR is related to https://github.com/src-d/engine/issues/132***

- I tested the changes and I could see how the siva-files aren't remove when the skipCleanup option is set to true while they were always removed before regardless the option were set or not.

***What was happening***
The `skinCleanup` K-V configuration was being set in the [SparkConf](https://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.SparkConf) object which is the configuration for the [SparkContext](https://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.SparkContext). It is in charge to manage the cluster's configuration stuffs. The `SparkConf` is given and cloned to the `SparkContext` when it is instantiated, so after that the `SparkConf` can't be used to add and remove additional configuration dynamically because they wouldn't actually tracked.

On the other hand, [SQLContext]() and [SparkSession]() have both configuration maps that we can use for get and set our own configuration things dynamically.

`SQLContext` is the old entry point to work with the SQL api things, meanwhile `SparkSession` is the new recommended way to do that. `SparkSession` encapsulate the `SparkContext` and the `SQLContext` for backward compatibility.

Although `sqlContext.getAllConfs`(the configuration Map(String, String)) and the `spark.conf.getAll` (the `SparkSession` configuration Map(String, String)) are different objects:

```scala
scala> spark.sqlContext.getAllConfs eq spark.conf.getAll
res3: Boolean = false
```

Both maps track the same K-V pairs:

```scala
scala> spark.sqlContext.getAllConfs == spark.conf.getAll
res4: Boolean = true
```

So when an option is set in one of them, automatically is added to the other too:

```scala
scala> spark.conf.set("spark.tech.sourced.engine.test.key", "test-value")

scala> spark.conf.get("spark.tech.sourced.engine.test.key")
res6: String = test-value

scala> spark.sqlContext.getConf("spark.tech.sourced.engine.test.key")
res8: String = test-value

scala> spark.sqlContext.getAllConfs eq spark.conf.getAll
res9: Boolean = false

scala> spark.sqlContext.getAllConfs == spark.conf.getAll
res10: Boolean = true
```

This is the way we should handle our own configuration things, not using the `SparkConf` object.

As it showed before `SQLContext` and `SparkSession` can be used for this purpose, but in my opinion, it's better use it through the `SparkSession` configuration since is the object "officially" recommended to be the entry point to the SQL API.

### DIFF
--- a/src/main/scala/tech/sourced/engine/DefaultSource.scala
+++ b/src/main/scala/tech/sourced/engine/DefaultSource.scala
@@ -66,8 +66,8 @@ case class GitRelation(sqlContext: SQLContext,
 
   private val localPath: String = UtilsWrapper.getLocalDir(sqlContext.sparkContext.getConf)
   private val path: String = sqlContext.getConf(repositoriesPathKey)
-  private val skipCleanup: Boolean = sqlContext.sparkContext.getConf
-    .getBoolean(skipCleanupKey, defaultValue = false)
+  private val skipCleanup: Boolean = sqlContext.sparkSession.conf.
+    get(skipCleanupKey, default = "false").toBoolean
 
   override def unhandledFilters(filters: Array[Filter]): Array[Filter] = {
     super.unhandledFilters(filters)
--- a/src/main/scala/tech/sourced/engine/Engine.scala
+++ b/src/main/scala/tech/sourced/engine/Engine.scala
@@ -157,7 +157,7 @@ class Engine(session: SparkSession) {
     * @return instance of the engine itself
     */
   def skipCleanup(skip: Boolean): Engine = {
-    session.sparkContext.getConf.set(skipCleanupKey, skip.toString)
+    session.conf.set(skipCleanupKey, skip)
     this
   }
 


bzz:
@manucorporat nice suggestion!
AFAIU that should still work with `spark-submit --conf`, right?

Would that make sense to make sure we use same configuration mechanism of SparkSession consistently in Engine? 

Asking, as quicks search shows https://github.com/src-d/engine/blob/dd8de9762c4e022859eb36feef1e39a78bbd628e/src/main/scala/tech/sourced/engine/udf/QueryXPathUDF.scala#L14

mcarmonaa:
@bzz all the configurations key=value loaded from `spark-defaults.conf` or from the command flag `--conf` are stored in the `SparkConf` object as wel as in the `SparkSession/SQLContext` configuration maps, so it could be accessed from all the "configuration stores".

Said this, I aggre that would be nice to use the `SparkSession` as a mechanism to handle all configuration things to do it consistently through all the engine code.

I can do another PR with these modifications if all the team agree too.

What do you think @ajnavarro @bzz @erizocosmico?

PR #152, opened 2017-11-02
Improve Jobs speed while reading siva files.

ajnavarro:
Remove unnecesary syncronized signature in a method and refactor the content a bit to make the Repository creation faster.

Executing:

spark.time(engine.getRepositories.getReferences.count())

on a 1000 siva files sample, local machine with master=local[8]

Before: 3.88253333 min
After: 2.23136667 min

### DIFF
--- a/src/main/scala/tech/sourced/engine/provider/RepositoryProvider.scala
+++ b/src/main/scala/tech/sourced/engine/provider/RepositoryProvider.scala
@@ -2,7 +2,6 @@ package tech.sourced.engine.provider
 
 import java.io.File
 import java.nio.file.Paths
-import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicInteger
 
 import org.apache.commons.io.FileUtils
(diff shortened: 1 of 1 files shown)


mcarmonaa:
Trait's method `concurrent.Map.getOrElseUpdate` doesn't ensure that the function will be evaluate just once or the result will be inserted, so maybe it would be nice leave a comment on the `repositories` and `repoRefCounts`declaration telling that `HashConcurrentMap` ensure these things.

If we need to change in the future the kind of map used, we won't forget to pay attention to that little detail.

ajnavarro:
@mcarmonaa I changed to a TrieMap as a concurrent Map implementation. It overrides the getOrElseUpdate implementation to make them thread safe.

mcarmonaa:
From the `TrieMap.getOrElseUpdate`s scaladoc:

```
def getOrElseUpdate(key: A, op: ⇒ B): B

Note: This method will invoke op at most once.
However, op may be invoked without the result being added to the map if a concurrent process is also trying to add a value corresponding to the same key k
```

I think `ConcurrentHashMap` was ok since it's implemented with performance as a goal and I think it ensures the things that the trait `concurrent.Map` delegates to the specific class implementation of the method.

I was talking about just leave a comment for the future to look for these details in the class implementation, if we need change the map, because the trait relies these things on it.

Anyway it's just a suggestion :+1:

ajnavarro:
Closing as invalid implementation.

Issue #151, opened 2017-10-31
ArrayIndexOutOfBoundsException in RepositoryProvider


warenlg:
In PySpark, 
```
from sourced.engine import Engine
engine.repositories.references.head_ref.files.extract_uasts().select("file_hash", "path", "uast").show()
```
resulted in 
```
[Stage 0:>                                                          (0 + 5) / 5]17/10/31 17:41:16 ERROR Executor: Exception in task 3.0 in stage 0.0 (TID 3)
java.lang.ArrayIndexOutOfBoundsException
	at java.lang.System.arraycopy(Native Method)
	at sun.security.provider.DigestBase.engineUpdate(DigestBase.java:114)
	at sun.security.provider.MD5.implDigest(MD5.java:101)
	at sun.security.provider.DigestBase.engineDigest(DigestBase.java:181)
	at sun.security.provider.DigestBase.engineDigest(DigestBase.java:160)
	at java.security.MessageDigest$Delegate.engineDigest(MessageDigest.java:592)
	at java.security.MessageDigest.digest(MessageDigest.java:365)
	at java.security.MessageDigest.digest(MessageDigest.java:411)
	at tech.sourced.engine.util.MD5Gen$.str(MD5Gen.scala:22)
	at tech.sourced.engine.provider.RepositoryProvider.genRepository(RepositoryProvider.scala:126)
	at tech.sourced.engine.provider.RepositoryProvider.get(RepositoryProvider.scala:57)
	at tech.sourced.engine.provider.RepositoryProvider.get(RepositoryProvider.scala:83)
	at tech.sourced.engine.GitRelation$$anonfun$buildScan$1.apply(DefaultSource.scala:87)
	at tech.sourced.engine.GitRelation$$anonfun$buildScan$1.apply(DefaultSource.scala:85)
	at scala.collection.Iterator$$anon$12.nextCur(Iterator.scala:434)
	at scala.collection.Iterator$$anon$12.hasNext(Iterator.scala:440)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIterator.agg_doAggregateWithKeys$(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIterator.processNext(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
```

[...]

PR #153, opened 2017-11-02
Create a new MessageDigest instance per call

ajnavarro:
The instance takes 2ms to be created and MessageDigest is not thread safe.


ajnavarro:
should fix https://github.com/src-d/engine/issues/151

erizocosmico:
Shall we merge and release this?

PR #159, opened 2017-11-03
Use an object pool to mange Repositories instances

ajnavarro:
This change manage to process ~1200 siva files repositories in less time:

Executed code: 
```scala
spark.time(engine.getRepositories.getReferences.count())
```

Using one core:
- Before: ~5 min
- Now: ~4,5 min

Using 8 cores:
- Before: ~4 min
- Now: ~2 min

### DIFF
--- a/src/main/scala/tech/sourced/engine/provider/RepositoryProvider.scala
+++ b/src/main/scala/tech/sourced/engine/provider/RepositoryProvider.scala
@@ -2,12 +2,12 @@ package tech.sourced.engine.provider
 
 import java.io.File
 import java.nio.file.Paths
-import java.util.concurrent.ConcurrentHashMap
-import java.util.concurrent.atomic.AtomicInteger
 
 import org.apache.commons.io.FileUtils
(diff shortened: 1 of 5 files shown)


src/main/scala/tech/sourced/engine/provider/RepositoryProvider.scala
```diff
@@ -24,19 +23,24 @@ import scala.collection.concurrent
   * @param localPath   Local path where siva files are.
   * @param skipCleanup Skip deleting files after they reference count of a repository gets to 0.
   */
-class RepositoryProvider(val localPath: String, val skipCleanup: Boolean = false) extends Logging {
+class RepositoryProvider(val localPath: String, val skipCleanup: Boolean = false)
+  extends Logging {
+  private val repositoryObjectFactory =
+    new RepositoryObjectFactory(localPath, skipCleanup)
+  private val genericKeyedObjectPool =
```
erizocosmico: s/genericKeyedObjectPool/repoPool/, what do you think? The `genericKeyedObject` part does not give much info about what it does

PR #163, opened 2017-11-07
better error message when engine is missing from pyspark

### LINKED ISSUE #160: Instruction how to launch engine from python
Hi,
I could launch engine with help of @ajnavarro but it's not so obvious.
We had to copy `engine.jar` to `pyspark/jars`.
I think it's not the best way (ultrahardcode), if you know better way - I will be glad to use it.
After copying you can create engine with:
```python
import sys

pyspark_path = "/path/to/pyspark/python"
sys.path.append(pyspark_path)

from bblfsh import BblfshClient, filter
from pyspark.sql import SparkSession
from sourced.engine import Engine

siva_folder = "/path/to/siva-files"
spark = SparkSession.builder.appName("test").master("local[*]").getOrCreate()
engine = Engine(spark, siva_folder)
```

### LINKED ISSUE #162: Better error message if jar is not loaded and update docs
- When jar is not loaded, Py4J throws an exception like `TypeError: 'JavaPackage' object is not callable`, we should catch that exception and throw another saying that `jar` with engine is not loaded and available in the JVM
- Update README so it reflects that jar has to be manually loaded when using PySpark

erizocosmico:
Closes #160 
Closes #162 

As flaky as this might seem, there is nothing in the Py4J API to check if something imported does actually exist or not and all it throws is a `TypeError` so that's our best guess.

### DIFF
--- a/python/sourced/engine/engine.py
+++ b/python/sourced/engine/engine.py
@@ -28,7 +28,15 @@ class Engine(object):
         self.__jvm = self.session.sparkContext._gateway.jvm
         java_import(self.__jvm, 'tech.sourced.engine.Engine')
         java_import(self.__jvm, 'tech.sourced.engine.package$')
-        self.__engine = self.__jvm.tech.sourced.engine.Engine.apply(self.__jsparkSession, repos_path)
+
+        try:
+            self.__engine = self.__jvm.tech.sourced.engine.Engine.apply(self.__jsparkSession, repos_path)
+        except TypeError as e:
+            if 'JavaPackage' in e.message:
+                raise Exception("package \"tech.sourced:engine:<version>\" cannot be found. Please, provide a jar with the package or install the package using --packages")
+            else:
+                raise e
+
         if skip_cleanup:
             self.__engine.skipCleanup(True)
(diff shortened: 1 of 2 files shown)


python/sourced/engine/engine.py
```diff
@@ -28,7 +28,15 @@ def __init__(self, session, repos_path, skip_cleanup=False):
         self.__jvm = self.session.sparkContext._gateway.jvm
         java_import(self.__jvm, 'tech.sourced.engine.Engine')
         java_import(self.__jvm, 'tech.sourced.engine.package$')
-        self.__engine = self.__jvm.tech.sourced.engine.Engine.apply(self.__jsparkSession, repos_path)
+
+        try:
+            self.__engine = self.__jvm.tech.sourced.engine.Engine.apply(self.__jsparkSession, repos_path)
+        except TypeError as e:
+            if 'JavaPackage' in e.message:
```
mcarmonaa: Does `JavaPackage` only appear in the exception we are trying to catch?

erizocosmico: In this particular line, only that `TypeError` can contain `JavaPackage`. At least that I can think of

PR #187, opened 2017-11-14
Refactor UDFs and Bblfsh code

mcarmonaa:
Changes:

- All UDFs code has been modified to use them in an uniformly way. It tries to improve the 
understandability and maintainability of the code keeping a common interface and usage style. You can call an UDF just with `myCoolUDF(col1, col2, col3)(implicit SparkSession)`. The UDFs are `case objects` that extends `CustomUDF`. All of them implements `name`, `function` and `apply`. 

- All bblfsh code has been gathered under the `Bbflsh` object, so configuration and client is wrapped and used directly from here. It avoids to have `SparkSession`s, `Bblfsh.Config`s and `Bblfsh.Client` mixed with the rest of the engine code. To get configuration just `Bblfsh.getConfig`and for extractUast and filter nodes just use `Bblfsh.extractUAST` and `BBlfsh.filter`

- Bblfsh errors are logged too:
```scala
scala> engine.getRepositories.getHEAD.getCommits.getFirstReferenceCommit.getFiles.extractUASTs.show
17/11/14 13:14:37 WARN Bblfsh: FATAL app/src/main/res/values/strings.xml: unexpected error: runtime failure: missing driver for language "xml"
17/11/14 13:14:37 WARN Bblfsh: FATAL app/src/main/res/values/strings.xml: unexpected error: runtime failure: missing driver for language "xml"
17/11/14 13:14:37 WARN Bblfsh: FATAL settings.gradle: unexpected error: runtime failure: missing driver for language "gradle"
17/11/14 13:14:37 WARN Bblfsh: FATAL settings.gradle: unexpected error: runtime failure: missing driver for language "gradle"
17/11/14 13:14:37 WARN Bblfsh: FATAL settings.gradle: unexpected error: runtime failure: missing driver for language "gradle"
17/11/14 13:14:37 WARN Bblfsh: FATAL settings.gradle: unexpected error: runtime failure: missing driver for language "gradle"
17/11/14 13:14:37 WARN Bblfsh: FATAL app/src/main/res/values/dimens.xml: unexpected error: runtime failure: missing driver for language "xml"
17/11/14 13:14:37 WARN Bblfsh: FATAL app/src/main/res/values/dimens.xml: unexpected error: runtime failure: missing driver for language "xml"
17/11/14 13:14:37 WARN Bblfsh: FATAL app/proguard-rules.txt: unexpected error: runtime failure: missing driver for language "text"
17/11/14 13:14:37 WARN Bblfsh: FATAL app/proguard-rules.txt: unexpected error: runtime failure: missing driver for language "text"
17/11/14 13:14:37 WARN Bblfsh: FATAL Google Image Downloader/create_dir.py~: unexpected error: runtime failure: missing driver for language ""
17/11/14 13:14:37 WARN Bblfsh: FATAL Google Image Downloader/create_dir.py~: unexpected error: runtime failure: missing driver for language ""
17/11/14 13:14:37 WARN Bblfsh: FATAL app/src/main/res/menu/main.xml: unexpected error: runtime failure: missing driver for language "xml"
17/11/14 13:14:37 WARN Bblfsh: FATAL app/src/main/res/menu/main.xml: unexpected error: runtime failure: missing driver for language "xml"
17/11/14 13:14:37 WARN Bblfsh: FATAL app/.gitignore: unexpected error: runtime failure: missing driver for language ""
17/11/14 13:14:37 WARN Bblfsh: FATAL app/.gitignore: unexpected error: runtime failure: missing driver for language ""
17/11/14 13:14:37 WARN Bblfsh: FATAL app/src/main/res/menu/login.xml: unexpected error: runtime failure: missing driver for language "xml"
17/11/14 13:14:38 WARN Bblfsh: FATAL app/src/main/res/menu/login.xml: unexpected error: runtime failure: missing driver for language "xml"
17/11/14 13:14:38 WARN Bblfsh: FATAL gradlew.bat: unexpected error: runtime failure: missing driver for language "batchfile"
17/11/14 13:14:38 WARN Bblfsh: FATAL gradlew.bat: unexpected error: runtime failure: missing driver for language "batchfile"
```

### DIFF
--- a/src/main/scala/tech/sourced/engine/Engine.scala
+++ b/src/main/scala/tech/sourced/engine/Engine.scala
@@ -185,5 +185,4 @@ object Engine {
     new Engine(session)
       .setRepositoriesPath(repositoriesPath)
   }
-
 }
--- a/src/main/scala/tech/sourced/engine/Engine.scala
+++ b/src/main/scala/tech/sourced/engine/Engine.scala
@@ -185,4 +185,5 @@ object Engine {
     new Engine(session)
       .setRepositoriesPath(repositoriesPath)
   }
+
 }
--- a/src/main/scala/tech/sourced/engine/package.scala
+++ b/src/main/scala/tech/sourced/engine/package.scala
@@ -2,8 +2,9 @@ package tech.sourced
 
 import gopkg.in.bblfsh.sdk.v1.uast.generated.Node
 import org.apache.spark.SparkException
-import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.sql.{Column, DataFrame, SparkSession}
 import tech.sourced.engine.udf._
+import tech.sourced.engine.util.Bblfsh
 
 /**
   * Provides the [[tech.sourced.engine.Engine]] class, which is the main entry point
@@ -57,10 +58,12 @@ package object engine {
     def registerUDFs(): Unit = {
       SessionFunctions.UDFtoRegister.foreach(customUDF => session.udf.register(
         customUDF.name,
-        customUDF.function(session)
+        customUDF.function
       ))
     }
 
+    Bblfsh.setConfig(session)
+    session.sparkContext.broadcast(Bblfsh.getConfig())
   }
 
   /**
@@ -227,10 +230,7 @@ package object engine {
       */
     def classifyLanguages: DataFrame = {
       checkCols(df, "is_binary", "path", "content")
-      df.withColumn(
-        "lang",
-        ClassifyLanguagesUDF.function(df.sparkSession)('is_binary, 'path, 'content)
-      )
+      df.withColumn("lang", ClassifyLanguagesUDF('is_binary, 'path, 'content))
     }
 
     /**
@@ -247,14 +247,13 @@ package object engine {
       */
     def extractUASTs(): DataFrame = {
       checkCols(df, "path", "content")
-      if (df.columns.contains("lang")) {
-        df.withColumn(
-          "uast",
-          ExtractUASTsUDF.functionWithLang(df.sparkSession)('path, 'content, 'lang)
-        )
+      val lang: Column = if (df.columns.contains("lang")) {
+        df("lang")
       } else {
-        df.withColumn("uast", ExtractUASTsUDF.function(df.sparkSession)('path, 'content))
+        null
       }
+
+      df.withColumn("uast", ExtractUASTsUDF('path, 'content, lang))
     }
 
     /**
@@ -295,10 +294,9 @@ package object engine {
         throw new SparkException(s"DataFrame already contains a column named $outputColumn")
       }
 
-      df.withColumn(
-        outputColumn,
-        QueryXPathUDF(df.sparkSession, query)(df(queryColumn))
-      )
+      import org.apache.spark.sql.functions.lit
+      val queryB = df.sparkSession.sparkContext.broadcast(query)
+      df.withColumn(outputColumn, QueryXPathUDF(df(queryColumn), lit(queryB.value)))
     }
 
     /**
@@ -321,7 +319,7 @@ package object engine {
         throw new SparkException(s"DataFrame already contains a column named $outputColumn")
       }
 
-      df.withColumn(outputColumn, ExtractTokensUDF()(df(queryColumn)))
+      df.withColumn(outputColumn, ExtractTokensUDF(df(queryColumn)))
     }
 
   }
@@ -364,8 +362,10 @@ package object engine {
       */
     val UDFtoRegister: List[CustomUDF] = List(
       ClassifyLanguagesUDF,
-      ExtractUASTsUDF,
-      QueryXPathUDF
+      ExtractUASTsWithoutLangUDF,
+      ExtractUASTsWithLangUDF,
+      QueryXPathUDF,
+      ExtractTokensUDF
     )
   }
 
--- a/src/main/scala/tech/sourced/engine/package.scala
+++ b/src/main/scala/tech/sourced/engine/package.scala
@@ -4,7 +4,6 @@ import gopkg.in.bblfsh.sdk.v1.uast.generated.Node
 import org.apache.spark.SparkException
 import org.apache.spark.sql.{Column, DataFrame, SparkSession}
 import tech.sourced.engine.udf._
-import tech.sourced.engine.util.Bblfsh
 
 /**
   * Provides the [[tech.sourced.engine.Engine]] class, which is the main entry point
@@ -58,12 +57,10 @@ package object engine {
     def registerUDFs(): Unit = {
       SessionFunctions.UDFtoRegister.foreach(customUDF => session.udf.register(
         customUDF.name,
-        customUDF.function
+        customUDF.function(session)
       ))
     }
 
-    Bblfsh.setConfig(session)
-    session.sparkContext.broadcast(Bblfsh.getConfig())
   }
 
   /**
@@ -84,6 +81,8 @@ package object engine {
   implicit class EngineDataFrame(df: DataFrame) {
 
     import df.sparkSession.implicits._
+
+    implicit val session = df.sparkSession
 
     /**
       * Returns a new [[org.apache.spark.sql.DataFrame]] with the product of joining the
--- a/src/main/scala/tech/sourced/engine/package.scala
+++ b/src/main/scala/tech/sourced/engine/package.scala
@@ -295,8 +295,7 @@ package object engine {
       }
 
       import org.apache.spark.sql.functions.lit
-      val queryB = df.sparkSession.sparkContext.broadcast(query)
-      df.withColumn(outputColumn, QueryXPathUDF(df(queryColumn), lit(queryB.value)))
+      df.withColumn(outputColumn, QueryXPathUDF(df(queryColumn), lit(query)))
     }
 
     /**
--- a/src/main/scala/tech/sourced/engine/udf/ClassifyLanguagesUDF.scala
+++ b/src/main/scala/tech/sourced/engine/udf/ClassifyLanguagesUDF.scala
@@ -17,6 +17,14 @@ case object ClassifyLanguagesUDF extends CustomUDF {
     function()(isBinary, path, content)
   }
 
+  /**
+    * Gets the language of the given file and returns the guessed language or none.
+    *
+    * @param isBinary whether it's a binary file or not
+    * @param path     file path
+    * @param content  file content
+    * @return `None` if no language could be guessed, `Some(language)` otherwise.
+    */
   private def getLanguage(isBinary: Boolean, path: String, content: Array[Byte]): Option[String] = {
     if (isBinary) {
       None
--- a/src/main/scala/tech/sourced/engine/udf/ClassifyLanguagesUDF.scala
+++ b/src/main/scala/tech/sourced/engine/udf/ClassifyLanguagesUDF.scala
@@ -1,23 +1,20 @@
 package tech.sourced.engine.udf
 
-import org.apache.spark.sql.Column
 import org.apache.spark.sql.expressions.UserDefinedFunction
 import org.apache.spark.sql.functions.udf
+import org.apache.spark.sql.{Column, SparkSession}
 import tech.sourced.enry.Enry
 
-/**
-  * User defined function to guess languages of files.
-  */
+/** User defined function to guess languages of files. */
 case object ClassifyLanguagesUDF extends CustomUDF {
 
   override val name = "classifyLanguages"
 
-  override def function: UserDefinedFunction = {
+  override def function(session: SparkSession = null): UserDefinedFunction =
     udf[Option[String], Boolean, String, Array[Byte]](getLanguage)
-  }
 
   def apply(isBinary: Column, path: Column, content: Column): Column = {
-    function(isBinary, path, content)
+    function()(isBinary, path, content)
   }
 
   private def getLanguage(isBinary: Boolean, path: String, content: Array[Byte]): Option[String] = {
--- a/src/main/scala/tech/sourced/engine/udf/ClassifyLanguagesUDF.scala
+++ b/src/main/scala/tech/sourced/engine/udf/ClassifyLanguagesUDF.scala
@@ -1,6 +1,6 @@
 package tech.sourced.engine.udf
 
-import org.apache.spark.sql.SparkSession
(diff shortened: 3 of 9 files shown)


src/main/scala/tech/sourced/engine/package.scala
```diff
@@ -295,10 +294,9 @@ package object engine {
         throw new SparkException(s"DataFrame already contains a column named $outputColumn")
       }
 
-      df.withColumn(
-        outputColumn,
-        QueryXPathUDF(df.sparkSession, query)(df(queryColumn))
-      )
+      import org.apache.spark.sql.functions.lit
+      val queryB = df.sparkSession.sparkContext.broadcast(query)
```
ajnavarro: You don't need to broadcast values to UDF definitions

src/main/scala/tech/sourced/engine/package.scala
```diff
@@ -57,10 +58,12 @@ package object engine {
     def registerUDFs(): Unit = {
       SessionFunctions.UDFtoRegister.foreach(customUDF => session.udf.register(
         customUDF.name,
-        customUDF.function(session)
+        customUDF.function
       ))
     }
 
+    Bblfsh.setConfig(session)
+    session.sparkContext.broadcast(Bblfsh.getConfig())
```
ajnavarro: That is doing something?

src/main/scala/tech/sourced/engine/udf/ExtractUASTsUDF.scala
```diff
@@ -1,118 +1,76 @@
...
+  */
+object ExtractUASTsUDF {
+
+  def apply(path: Column,
+            content: Column,
+            lang: Column = null): Column = {
+    if (lang == null) {
+      ExtractUASTsWithoutLangUDF(path, content)
+    } else {
+      ExtractUASTsWithLangUDF.function(path, content, lang)
```
ajnavarro: Why on `ExtractUASTsWithLangUDF` we need to call to function method and in `ExtractUASTsWithoutLangUDF` not?

src/main/scala/tech/sourced/engine/util/Bblfsh.scala
```diff
@@ -17,34 +27,86 @@ object Bblfsh {
...
+    */
+  def extractUAST(path: String,
+                  content: Array[Byte],
+                  lang: String): Seq[Array[Byte]] = {
+
+    //FIXME(bzz): not everything is UTF-8 encoded :/
+
+    // if lang == null, it hasn't been classified yet
+    // so rely on bblfsh to guess this file's language
+    if (lang != null && !supportedLangs.contains(lang.toLowerCase())) {
```
ajnavarro: supportedLangs are in lowercase too?

src/main/scala/tech/sourced/engine/util/Bblfsh.scala
```diff
@@ -17,34 +27,86 @@ object Bblfsh {
...
+    } else {
+      config
+    }
+  }
+
+  private def getClient(): BblfshClient = synchronized {
     if (client == null) {
-      client = BblfshClient(config.host, config.port)
+      val Config(host, port) = getConfig()
+      client = BblfshClient(host, port)
```
ajnavarro: This client will always have the same default configuration on remote workers. we need to broadcast configuration from the driver to the workers. I cannot see where we are doing this.

src/main/scala/tech/sourced/engine/package.scala
mcarmonaa: Nope, let me move the bblfsh broadcasting configuration to the `Bblfsh` object

src/main/scala/tech/sourced/engine/package.scala
mcarmonaa: We were doing it before in the [UDF wrapper object](https://github.com/src-d/engine/blob/master/src/main/scala/tech/sourced/engine/udf/QueryXPathUDF.scala#L20)

I'll remove it

src/main/scala/tech/sourced/engine/udf/ExtractUASTsUDF.scala
mcarmonaa: We don't need, I forgot changed it

src/main/scala/tech/sourced/engine/util/Bblfsh.scala
mcarmonaa: This is related with your first comment, let' modify this

src/main/scala/tech/sourced/engine/util/Bblfsh.scala
mcarmonaa: Yes, they are `val supportedLangs = Set("python", "java")`

ajnavarro: Now thinking about this Set with the supported languages, Should we delete it? I mean, If bblfsh adds more languages and the user didn't precalculate languages using Enry, thew will not be able to send to bblfsh files with the new supported language.

mcarmonaa:
@ajnavarro now the bblfsh configuration is broadcast from `object Engine` when `apply` is used.

In this way, bblfsh utils is agnostic of `SparkSessions` and `Broadcast` variables.

src/main/scala/tech/sourced/engine/Engine.scala
```diff
@@ -182,7 +183,19 @@ object Engine {
...
   }
+
+  private def spreadBblfshConfig(session: SparkSession): Unit = {
+    val host = session.conf.get(bblfshHostKey, Bblfsh.defaultHost)
+    val port = session.conf.get(bblfshPortKey, Bblfsh.defaultPort.toString).toInt
+
+    val hostB = session.sparkContext.broadcast(host)
+    val portB = session.sparkContext.broadcast(port)
+
+    Bblfsh.setConfig(hostB.value, portB.value)
```
ajnavarro: This wont work. You are setting the configuration just to the local JVM. You should add as configuration the Broadcast value, not the value inside the broadcast itself.

src/main/scala/tech/sourced/engine/util/Bblfsh.scala
mcarmonaa: see [comment](https://github.com/src-d/engine/pull/187#issuecomment-344534977)

src/main/scala/tech/sourced/engine/util/Bblfsh.scala
mcarmonaa: see [comment](https://github.com/src-d/engine/pull/187#issuecomment-344534977)

erizocosmico: This should be delegated to bblfsh imho

src/main/scala/tech/sourced/engine/Engine.scala
mcarmonaa: Ok, so something like `Bblfsh.setConfig(hostB, portB)` and left the `Bblfsh.setConfig` implementation extract the values, right?

erizocosmico: Are you sure this works? Because even if the values of the host and port have been broadcasted, the Bblfsh object has not been mutated elsewhere

erizocosmico: Nope, that's mutating the local object as well. That's why before it was the way it was done

mcarmonaa: I see, thanks for the explanation @erizocosmico!

Because all of this, I wonder if it's actually needed broadcast the configuration then. I mean, the only purpose of this configuration is get a bblfsh client, and in the current implementation only one client is instantiated to reuse it purposes. So, isn't the configuration just sent once?

src/main/scala/tech/sourced/engine/util/Bblfsh.scala
mcarmonaa: We already had a Set to don't send `markdown` neither `text` as optimization. Should I come back to this approach? Or just not exclude any language?

src/main/scala/tech/sourced/engine/Engine.scala
erizocosmico: But you don't have the client instantiated in other nodes. So yes, you need to send it.

If you do a setConfig in node A, node B still has a `null` `client`, they run on completely different JVMs, AFAIK.

In local mode, of course, this works, because they share the same JVM.

src/main/scala/tech/sourced/engine/util/Bblfsh.scala
erizocosmico: Things we know will never have a driver, such as markdown and text, make sense to optimize. But languages that might have it in the future it's better to delegate to bblfsh, otherwise we must maintain that list and people using engine of version X will need to upgrade if they want to use language Y that was not supported back then.

src/main/scala/tech/sourced/engine/Engine.scala
mcarmonaa: In your example, node B with a `null` `client` would try to instantiate it, so it would request the configuration to the driver and the next time it will already have a `client`.

So the configuration would be sent once per node as well as if it's broadcast. Do I missing something?

Sorry for the mess and thank you for help me to understand this `Spark` related 
things!

src/main/scala/tech/sourced/engine/util/Bblfsh.scala
mcarmonaa: Ok, let me change it! :+1:

src/main/scala/tech/sourced/engine/Engine.scala
erizocosmico: `Engine` is not instantiated in all nodes, and thus, `setConfig` is not called in every node. Send does not mean, values will be available in every node if you store them somewhere, send means, I broadcast this value and give you a reference to it that is serializable, so in every task that will be sent to other nodes you will retrieve the same broadcasted value using that reference.
That's why the only why to access a broadcasted thing is having the broadcasted variable in scope.

src/main/scala/tech/sourced/engine/util/Bblfsh.scala
```diff
@@ -1,50 +1,130 @@
...
     * Key used for the option to specify the host of the bblfsh grpc service.
     */
-  val hostKey = "spark.tech.sourced.bblfsh.grpc.host"
+  val bblfshHostKey = "spark.tech.sourced.bblfsh.grpc.host"
 
   /**
     * Key used for the option to specify the port of the bblfsh grpc service.
     */
-  val portKey = "spark.tech.sourced.bblfsh.grpc.port"
+  val bblfshPortKey = "spark.tech.sourced.bblfsh.grpc.port"
```
erizocosmico: It's accessed as `Bblfsh.portKey`, the `bblfsh` prefix is kind of redundant given the scope of this property

mcarmonaa:
@ajnavarro @erizocosmico thanks for the explanations!

Eventually, I moved the broadcast bblfsh configuration code to the `Bblfsh.getClient` method which is used from the udfs, so now the configuration should be broadcast properly I guess.

Is it correct?

src/main/scala/tech/sourced/engine/util/Bblfsh.scala
```diff
@@ -1,50 +1,130 @@
...
+    }
+
+    config
+  }
+
+  private def getClient(): BblfshClient = synchronized {
     if (client == null) {
-      client = BblfshClient(config.host, config.port)
+      val Config(host, port) = getConfig()
+      if (broadcastConf) {
```
erizocosmico: This still does not work on multiple nodes. You need to hold the reference to the `Broadcast` inside the serialized function that is going to be sent to the nodes to execute. This is only being changed in the node where `Engine` is initialized.

mcarmonaa: I'll return again the config to a broadcast variable inside a closure where the udf is defined

mcarmonaa:
Eventually, I back again to the approach of broadcast the bblfsh configuration variable from a closure.

I added to the UDF objects' `apply` method the session as an implicit parameter, so all of them can be used with `myCoolUDF(col1, col2)` making the `SparkSession` an implicit val in the scope.

src/main/scala/tech/sourced/engine/udf/ClassifyLanguagesUDF.scala
```diff
@@ -1,34 +1,29 @@
...
-  /**
-    * Gets the language of the given file and returns the guessed language or none.
-    *
-    * @param isBinary whether it's a binary file or not
-    * @param path     file path
-    * @param content  file content
-    * @return `None` if no language could be guessed, `Some(language)` otherwise.
-    */
-  def getLanguage(isBinary: Boolean, path: String, content: Array[Byte]): Option[String] =
+  private def getLanguage(isBinary: Boolean, path: String, content: Array[Byte]): Option[String] = {
```
ajnavarro: why did you remove existing documentation?

src/main/scala/tech/sourced/engine/udf/CustomUDF.scala
```diff
@@ -3,13 +3,13 @@ package tech.sourced.engine.udf
 import org.apache.spark.sql.SparkSession
```
ajnavarro: This file should not be modified

src/main/scala/tech/sourced/engine/Engine.scala
```diff
@@ -185,4 +185,5 @@ object Engine {
     new Engine(session)
       .setRepositoriesPath(repositoriesPath)
   }
+
```
ajnavarro: This file should not be modified.

src/main/scala/tech/sourced/engine/udf/ClassifyLanguagesUDF.scala
mcarmonaa: because now is private

ajnavarro: Documentation is always good, so, if it was already there, I think is worth it keep there, even if the signature of the method is now private.

mcarmonaa: Ok, I'll add it again, just keep in mind that documentation has to be maintained too :+1:

mcarmonaa:
@ajnavarro can I merge this PR?

ajnavarro:
LGTM

Issue #183, opened 2017-11-14
Check list of unhandledFilters to don't filter two times with all the existing filters.


ajnavarro:
In BaseRelation we have:
```
  /**
   * Returns the list of [[Filter]]s that this datasource may not be able to handle.
   * These returned [[Filter]]s will be evaluated by Spark SQL after data is output by a scan.
   * By default, this function will return all filters, as it is always safe to
   * double evaluate a [[Filter]]. However, specific implementations can override this function to
   * avoid double filtering when they are capable of processing a filter internally.
   *
   * @since 1.6.0
   */
  def unhandledFilters(filters: Array[Filter]): Array[Filter] = filters
```

we should override this method to return a list of already-not-supported filters on our datasource, to avoid filter two times per each filter operation.

mcarmonaa:
This method takes in `Filter`s, but since we are building engine on top of `SQL spark` our filters come up as `Expression`s so there are divergences between the filters `unhandledFilters` catches and the expressions we are receiving in the `buildScan` method.

Also spark performs optimizations which transform a lot the queries, so we can not be sure about what filters eventually we will be able to handle at sources level for the moment.

erizocosmico:
Should we close this, then? @mcarmonaa @ajnavarro

PR #193, opened 2017-11-15
[WIP] override BaseRelation's method unhandledFilter

mcarmonaa:
***For the moment this task is stopped.***
This PR is related to #183 

Comment: there is a test failing. I suspect that it comes because before override the method  our filters and spark filters were both applied.


mcarmonaa:
Closing this until metadata changes will be ready.

`unhandledFilters` method takes in `Filter`s but we are working with `Expressions` at the `Spark SQL` level. Spark makes its own optimizations so sometimes the we don't get the expressions we expected.

On the other hand, we are not handling our `CompiledFilters` in the iterators to make the right filtering at this level, and we don't know if we can do that properly right now.

It needs more research when we will have the different data sources defined and implemented.

Issue #177, opened 2017-11-13
NullPointerException if .persist() is called


vmarkovtsev:
This is what happens when I call `engine.repositories.persist(StorageLevel.DISK_ONLY)`:
```
17/11/13 10:58:23 WARN GitOptimizer: Join cannot be optimized. It doesn't have GitRelations in both sides, or the Join type is not supported.
17/11/13 10:58:23 WARN GitOptimizer: Join cannot be optimized. Invalid node: Join Inner, (repository_id#24 = id#0)
:- Project [repository_id#24, name#25]
:  +- Filter (name#25 = refs/heads/HEAD)
:     +- Relation[repository_id#24,name#25,hash#26] GitRelation(org.apache.spark.sql.SparkSession@5e63c6e8,StructType(StructField(repository_id,StringType,false), StructField(name,StringType,false), StructField(hash,StringType,false)),None,Some(references))
+- Project [id#0]
   +- InMemoryRelation [id#0, urls#1, is_fork#2], true, 10000, StorageLevel(disk, 1 replicas)
         +- *Scan GitRelation(org.apache.spark.sql.SparkSession@5e63c6e8,StructType(StructField(id,StringType,false), StructField(urls,ArrayType(StringType,false),false), StructField(is_fork,BooleanType,true)),None,Some(repositories)) [id#0,urls#1,is_fork#2] ReadSchema: struct<id:string,urls:array<string>,is_fork:boolean>

17/11/13 10:58:23 WARN GitOptimizer: Join cannot be optimized. Invalid node: InMemoryRelation [id#0, urls#1, is_fork#2], true, 10000, StorageLevel(disk, 1 replicas)
   +- *Scan GitRelation(org.apache.spark.sql.SparkSession@5e63c6e8,StructType(StructField(id,StringType,false), StructField(urls,ArrayType(StringType,false),false), StructField(is_fork,BooleanType,true)),None,Some(repositories)) [id#0,urls#1,is_fork#2] ReadSchema: struct<id:string,urls:array<string>,is_fork:boolean>

17/11/13 10:58:23 WARN GitOptimizer: Join cannot be optimized. Invalid node: Join Inner, ((repository_id#24 = repository_id#66) && (reference_name#67 = name#25))
:- Project [repository_id#66, reference_name#67, hash#69]
:  +- Filter (reference_name#67 = refs/heads/HEAD)
[...]

PR #197, opened 2017-11-16
Fix NPE when persist DF to disk

mcarmonaa:
***This PR is related to #177***

The changes avoid to return an object to the pool which was already returned.

I don't know why, but when the DF is persisted to disk, the last repo in the RDD generated by the `BaseRelation.buildScan` is closed twice. Because of this, the NPE was thrown.

@ajnavarro @erizocosmico @jfontan do you know what is causing this behavior?


erizocosmico:
That's super weird. Spark bug maybe?

ajnavarro:
yep, really strange. Can we add a warning when that happens? to be able to check other cases that this is happening and handle them in the future.

mcarmonaa:
I added the warning, a comment where the close is called and a test for all storage levels.

PR #203, opened 2017-11-21
Moved storage level test to a new class

mcarmonaa:
Since persist is bound to a `SparkContext` and we reuse the same `SparkSession` for all the tests in `DefaultSourceSpec`, the storage level test has been located in a new class that use a different `SparkSession` to avoid erros in other tests that are not related.

### DIFF
--- a/src/test/scala/tech/sourced/engine/StorageLevelSpec.scala
+++ b/src/test/scala/tech/sourced/engine/StorageLevelSpec.scala
@@ -1,0 +1,37 @@
+package tech.sourced.engine
+
+import org.scalatest.{FlatSpec, Matchers}
+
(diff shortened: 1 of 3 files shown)


src/test/scala/tech/sourced/engine/StorageLevelSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+    )
+
+    storageLevels.foreach(level => {
+      val df = engine.getRepositories.persist(level)
+      df.count()
+      df.unpersist()
+    })
+  }
+
+  override protected def afterAll(): Unit = {
```
erizocosmico: I don't think we need to override afterAll if we're not doing anything special in it

PR #206, opened 2017-11-21
Support bucketed repository directories

### LINKED ISSUE #176: Be able to read rooted-repositories folder with bucketing

jfontan:
It also supports siva files at the first level as before. The only
cabeat is that it does not support more than one level of directories.

Fixes #176


.travis.yml
```diff
@@ -76,7 +76,7 @@ matrix:
             tags: true
 
 before_script:
-  - docker run --privileged -d -p 9432:9432 --name bblfsh bblfsh/server
+  - docker run --privileged -d -p 9432:9432 --name bblfsh bblfsh/bblfshd
```
mcarmonaa: You should install the driver after run the container, anyway maybe it would be better don't include this change in the PR and rebase on #205 instead.

jfontan: I'll do. Thanks!

jfontan:
I've dropped the bblfsh change but have not rebased the changes with #205 as it's still not in the main repo.

jfontan:
The changes are now rebased to master. Tests are passing.

erizocosmico:
:shipit: 🚢

PR #228, opened 2017-11-29
Added rule to attach source to attributes

mcarmonaa:
Before this PR, the `source` assignation as metadata to `AttributeReference` was done in the `SquashGitRelationJoin` so a code like this didn't work properly:

```scala
val repoDf = engine.getRepositories
repoDf.where("id='github.com/xiyou-linuxer/faq-xiyoulinux'").show
```

It was filtered by spark instead of us because the `source` it wasn't  set.

Now there is a new rule `AddSourceToAttributes` ensuring the `source` is set always.

### DIFF
--- a/src/main/scala/tech/sourced/engine/GitOptimizer.scala
+++ b/src/main/scala/tech/sourced/engine/GitOptimizer.scala
@@ -10,13 +10,38 @@ import org.apache.spark.sql.execution.datasources.LogicalRelation
 import org.apache.spark.sql.types._
 
 /**
+  * Rule to assign to an [[AttributeReference]] metadata to identify the table it belongs to.
+  */
+object AddSourceToAttributes extends Rule[LogicalPlan] {
+
+  /**
+    * SOURCE is the key used for attach metadata to [[AttributeReference]]s.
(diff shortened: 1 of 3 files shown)


src/main/scala/tech/sourced/engine/GitOptimizer.scala
```diff
@@ -9,14 +9,42 @@ import org.apache.spark.sql.catalyst.rules.Rule
...
+object AddSourceToAttributes extends Rule[LogicalPlan] {
+
+  /**
+    * SOURCE is the key used for attach metadata to [[AttributeReference]]s.
+    */
+  val SOURCE = "source"
+
+  /** @inheritdoc */
+  def apply(plan: LogicalPlan): LogicalPlan = plan transformUp {
+    case LogicalRelation(
```
jfontan: This indentation feels a bit weird.

mcarmonaa: Actually it does, the `case` line at some point was longer than 100 characters but it's not now. Let me write it again in just one line :+1:

PR #230, opened 2017-11-29
Store metadata in a SQLite database

erizocosmico:
Depends on #227 

Before you ask, the `is_fork` column is being replaced because after a lot of debugging figured out that nullable boolean (not sure if not nullable boolean) is not handled well in the SQLite jdbc driver so fuck it, nullable integer it is, we have to un-transform arrays so we might as well transform `is_fork` when reading from the db.

EDIT: not-nullable boolean does not work either. `BIT` is not handled at all in the driver. FML.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,4 +81,3 @@ before_script:
 
 script:
   - make travis-test
-  - make build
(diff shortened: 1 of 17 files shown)


erizocosmico:
Superseded by #232

Issue #244, opened 2017-12-04
java.nio.file.InvalidPathException: Malformed input or input contains unmappable characters


EgorBu:
Hi,
I tried to process many thousands of repositories provided by @bzz and it crashed with log `Malformed input or input contains unmappable characters` (more details below).
I attached `siva` file that allows reproducing the error.
[d920e267dcd75a3eb6da57bea9e4178db8891d9b.siva.zip](https://github.com/src-d/engine/files/1527626/d920e267dcd75a3eb6da57bea9e4178db8891d9b.siva.zip)

<details>

```
java.nio.file.InvalidPathException: Malformed input or input contains unmappable characters: /tmp/spark/spark-4a1b0b45-fcbf-438d-8f30-492f95df04e0/processing-repositories/382B3EA5A2BD600984AFC7924CC85C46/d920e267dcd75a3eb6da57bea9e4178db8891d9b.siva/refs/tags/???????????????/015f72dd-e755-7eb2-524a-90ba40bbfeaa
	at sun.nio.fs.UnixPath.encode(UnixPath.java:147)
	at sun.nio.fs.UnixPath.<init>(UnixPath.java:71)
	at sun.nio.fs.UnixFileSystem.getPath(UnixFileSystem.java:281)
	at java.nio.file.Paths.get(Paths.java:84)
	at tech.sourced.engine.provider.RepositoryObjectFactory$$anonfun$genSivaRepository$1.apply(RepositoryProvider.scala:209)
	at tech.sourced.engine.provider.RepositoryObjectFactory$$anonfun$genSivaRepository$1.apply(RepositoryProvider.scala:207)
	at scala.collection.Iterator$class.foreach(Iterator.scala:893)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1336)
	at scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
	at tech.sourced.engine.provider.RepositoryObjectFactory.genSivaRepository(RepositoryProvider.scala:207)
	at tech.sourced.engine.provider.RepositoryObjectFactory.create(RepositoryProvider.scala:92)
	at tech.sourced.engine.provider.RepositoryObjectFactory.create(RepositoryProvider.scala:86)
	at org.apache.commons.pool2.BaseKeyedPooledObjectFactory.makeObject(BaseKeyedPooledObjectFactory.java:60)
[...]

PR #257, opened 2017-12-12
Document file name encoding issue

mcarmonaa:
Added explanation about how and why set `LANG` environment variable to avoid encoding issues.

it comes from #244

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,12 @@ $ pyspark --packages "tech.sourced:engine:[version]"
 
 Run [bblfsh daemon](https://github.com/bblfsh/bblfshd). You can start it easily in a container following its [quick start guide](https://github.com/bblfsh/bblfshd#quick-start).
 
+If you run **engine** in an UNIX like environment, you should set the `LANG` variable properly:
+
+    export LANG="en_US.UTF-8"
+
+The rationale behind this is that UNIX file systems don't keep the encoding for each file name, they are just plain bytes,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,6 +46,12 @@ $ pyspark --packages "tech.sourced:engine:[version]"
 
 Run [bblfsh daemon](https://github.com/bblfsh/bblfshd). You can start it easily in a container following its [quick start guide](https://github.com/bblfsh/bblfshd#quick-start).
 
+If you run **engine** in a UNIX like environment, you should set the `LANG` variable properly:
```
erizocosmico: s/a/an/

README.md
```diff
@@ -46,6 +46,12 @@ $ pyspark --packages "tech.sourced:engine:[version]"
...
+
+    export LANG="en_US.UTF-8"
+
+The rationale behind this is because UNIX file systems don't keep the encoding for each file name, they are just plain bytes,
```
erizocosmico: /this is because/this is that/

bzz:
👍 

May be adding a symptoms could help people identify this case, it they stumbled up on it? Like 
"and thus `java.nio.file.InvalidPathException: Malformed input or input contains unmappable characters` would happen if utf filenames are read in C locale"

mcarmonaa:
As @bzz suggested, I added a few lines about what happens when the `LANG` environment variable is set wrong.

PR #263, opened 2017-12-14
engine: make getHEAD match also HEAD, rewrite remote refs

### LINKED ISSUE #255: getHEAD method is filtering by refs/heads/HEAD instead of HEAD in bare or standard repositories formats.
This might or might not be connected to #254 but on similar use case of using Engine on a local Git repository and trying to count number of commits, one gets unexpected result.

Steps to reproduces:
```
mkdir test-repo-root
cd test-repo-root; git clone git@github.com:bblfsh/client-scala.git; cd -

"${SPARK_HOME}"/bin/spark-shell --packages "tech.sourced:engine:0.1.12"
```

And then

```
import tech.sourced.engine._
val engine = Engine(spark, "./test-repo-root/")
engine.getRepositories.getHEAD.getCommits.getFirstReferenceCommit.count
```
results in `0`.

Expected results would be closer to
```
$ cd
$ git rev-list HEAD --count
125
```

erizocosmico:
Fixes #255

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,13 @@ These are all the supported formats at the moment:
 - `standard`: regular git repositories with a `.git` folder. Each in a folder of their own under the given repository path.
 - `bare`: git bare repositories. Each in a folder of their own under the given repository path.
 
+### Processing local repositories with the engine
+
+There are some design decisions that may surprise the user when processing local repositories, instead of siva files. This is the list of things you should take into account when doing so:
+
+- All local branches will belong to a repository whose id is `file://$REPOSITORY_PATH`. So, if you clone `https://github.com/foo/bar.git` at `/home/foo/bar`, you will see two repositories `file:///home/foo/bar` and `github.com/foo/bar`, even if you only have one.
+- Remote branches are transformed from `refs/remote/$REMOTE_NAME/$BRANCH_NAME` to `refs/heads/$BRANCH_NAME` as they will only belong to the repository id of their corresponding remote. So `refs/remote/origin/HEAD` becomes `refs/heads/HEAD`.
+
 # Playing around with **engine** on Jupyter
 
 You can launch our docker container which contains some Notebooks examples just running:
--- a/src/main/scala/tech/sourced/engine/package.scala
+++ b/src/main/scala/tech/sourced/engine/package.scala
@@ -238,6 +238,9 @@ package object engine {
       }
     }
 
+    private val HeadRef = "refs/heads/HEAD"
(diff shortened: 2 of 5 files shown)


src/main/scala/tech/sourced/engine/package.scala
```diff
@@ -205,6 +205,9 @@ package object engine {
       }
     }
 
+    private val headRef = "refs/heads/HEAD"
```
mcarmonaa: Should the name's first letter be capitalized as the other constants?

src/main/scala/tech/sourced/engine/package.scala
```diff
@@ -205,6 +205,9 @@ package object engine {
...
+    private val localHeadRef = "HEAD"
```
mcarmonaa: Same constant name style here

README.md
```diff
@@ -209,6 +209,13 @@ scala> engine.getRepositories.filter('id === "github.com/mawag/faq-xiyoulinux").
 
 ```
 
+### Processing local repositories with the engine
+
+There are some design decisions that may surprise the user when processing local repositories, instead of siva files. This is the list of things you should take into account when doing so:
+
+- All local branches will belong to a repository whose id is `file://$REPOSITORY_PATH`. So, if you clone `https://github.com/foo/bar.git` at `/home/foo/bar`, you will see two repositories `file:///home/foo/bar` and `github.com/foo/bar`, even if you only have one.
+- Remote branches are transformed from `refs/remote/$REMOTE_NAME/$BRANCH_NAME` to `refs/heads/$BRANCH_NAME` as they will only belong to the repository id of their corresponding remote. So `refs/remote/origin/HEAD` becomes `refs/heads/HEAD`.
```
mcarmonaa: Aren't there name collision issues with that?

erizocosmico: Nope, they belong to different repositories. If you have a local branch that's `refs/heads/develop` and a `refs/remote/foo/develop` the second one will belong to `github.com/foo/bar` and the first one to `file:///yada/yada`

erizocosmico:
@ajnavarro @jfontan @mcarmonaa perhaps we should add `is_remote` field to the `references` table? Since we're rewriting them, it could be useful to be able to filter by whether a ref is remote or not.

That way, local branches would be `false`, remote branches would be `true` and all siva branches would be `true` (because they all come from a remote, right?)

jfontan:
I don't see any reason why we shouldn't add that field. It should make some reference filters faster.

erizocosmico:
I will address that in another PR and merge this.

Issue #268, opened 2017-12-19
"AttributeError: 'TypeError' object has no attribute 'message'" when handling unavailable engine.jar


EgorBu:
Your issue may already be reported!
Please search on the [issue track](../) before creating one.

## Expected Behavior
Error while catching of exception when `engine.jar` is not provided.
It should print (according to idea in code) that `jar` is not available but fails.

## Current Behavior
When `engine.jar` is not available:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/usr/local/lib/python3.5/dist-packages/sourced/engine/engine.py in __init__(self, session, repos_path, skip_cleanup)
     32         try:
---> 33             self.__engine = self.__jvm.tech.sourced.engine.Engine.apply(self.__jsparkSession, repos_path)
     34         except TypeError as e:

TypeError: 'JavaPackage' object is not callable

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
<ipython-input-2-d4a516b1e3f7> in <module>()
      9 siva_folder = "/backup/siva/30k/"
     10 # siva_folder = "/data/siva/100-java/"
---> 11 engine = Engine(spark, siva_folder)

/usr/local/lib/python3.5/dist-packages/sourced/engine/engine.py in __init__(self, session, repos_path, skip_cleanup)
     33             self.__engine = self.__jvm.tech.sourced.engine.Engine.apply(self.__jsparkSession, repos_path)
     34         except TypeError as e:
---> 35             if 'JavaPackage' in e.message:
     36                 raise Exception("package \"tech.sourced:engine:<version>\" cannot be found. Please, provide a jar with the package or install the package using --packages")
     37             else:

AttributeError: 'TypeError' object has no attribute 'message'
```

## Possible Solution
One of the way to check is `e.args` 
```python
try:
    x = 1
    x()
except TypeError as e:
    print(e)
    print(dir(e))
    print(e.args)

    if "int" in e.args[0]:
        raise Exception("package \"tech.sourced:engine:<version>\" cannot be found. Please, provide a jar with the package or install the package using --packages")
```
and it gives:
```
'int' object is not callable
[...]

PR #269, opened 2017-12-19
Fix to "AttributeError: 'TypeError' object has no attribute 'message'" when handling unavailable engine.jar

EgorBu:
Fix https://github.com/src-d/engine/issues/268


EgorBu:
Made comment in issue: https://github.com/src-d/engine/issues/268#issuecomment-352744296
It should be a solution for both python 2 & 3.

PR #274, opened 2017-12-19
Add default `repos_format`

EgorBu:
Let's make default `repos_format` to avoid sending this parameter always (in our case we use `siva` files almost always) + back compatibility with the previous version of `engine`.

+ it will make documentation in this class correct.
Right now it doesn't have `repos_format` (https://github.com/src-d/engine/blob/master/python/sourced/engine/engine.py#L12~#L14):
```
    >>> from sourced.engine import Engine
    >>> repos_df = Engine(sparkSession, "/path/to/my/repositories").repositories
    >>> repos_df.show()
```

### DIFF
--- a/python/sourced/engine/engine.py
+++ b/python/sourced/engine/engine.py
@@ -24,7 +24,7 @@ class Engine(object):
     :type skip_cleanup: bool
     """
 
-    def __init__(self, session, repos_path, repos_format, skip_cleanup=False):
+    def __init__(self, session, repos_path, repos_format='siva', skip_cleanup=False):
         self.session = session
         self.__jsparkSession = session._jsparkSession
         self.session.conf.set('spark.tech.sourced.engine.repositories.path', repos_path)


ajnavarro:
Have a no default value for that is on purpose. Depending of that value, we are parsing the repository directory in different ways. Is not possible for us to guess which format are you using and manage correctly the processes parallelization in some cases, That's why we decide to change the format of the repositories to a mandatory parameter.

EgorBu:
I think it's not so critical because a lot of people will copy-paste code and this errors (wrong format) will appear a lot of times anyway. I believe that it's better to make some checks before or raise an error as early as possible.

EgorBu:
what do you think? @ajnavarro @erizocosmico @mcarmonaa @jfontan

ajnavarro:
If the people are copy-pasting code, and that code throw an error, they must try to understand the code that they are typing. For this, I consider that is easier to understand a mandatory parameter that you must have into your code, that a default parameter that can be easily missed when debugging the problem.

mcuadros:
What ever you decided, if you make something default please not `siva` make raw repositories.

erizocosmico:
I think default repository format makes sense, but as @mcuadros says, it should not be siva, it should be raw repositories, which is what everyone would expect while copy pasting stuff to try out examples. That way, raw repositories become kind of the default thing and siva files (and other formats) are an opt-in thing.

I mean, `Engine(spark, path, "standard")` is not really much more self-explanatory than `Engine(spark, path)`, as `standard` here might be kinda confusing in the context of someone that is trying examples by copy-pasting.

Perhaps `standard` -> `git` and `bare` -> `git-bare` would make this easier for newcomers, but still think having `standard` as default makes sense.

erizocosmico:
Any updates on this?

mcuadros:
Lets revisti this for engine v2

PR #278, opened 2017-12-21
Mute loggers on tests

mcarmonaa:
Supersed #231

### DIFF
--- a/src/test/resources/log4j.properties
+++ b/src/test/resources/log4j.properties
@@ -1,12 +1,12 @@
 # Define the root logger with appender
-log4j.rootCategory=WARN, console
+log4j.rootCategory=ERROR, console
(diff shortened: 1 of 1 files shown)


mcarmonaa:
@ajnavarro will not the changes in this config file affect to loggers always? I mean, change them programatically allow us to set off only for tests, and if needed, just commenting these lines we can turning on again for some quick debugging.

ajnavarro:
This file is only affecting all the logs on tests.

> and if any problem is detected running tests, just commenting these lines we can turning on again for some quick debugging

You can comment or change that file on debugging too

mcarmonaa:
@ajnavarro changes done on `src/test/resources/log4j.properties`. Much better now :+1:

Issue #273, opened 2017-12-19
[METADATA STORE] Add Python bindings to the Engine API


ajnavarro:
We need to add all the bindings that are missed to use the metadata store on the python side.

mcarmonaa:
Done in #282

PR #282, opened 2018-01-02
Added Python bindings for metadata sources

mcarmonaa:
This PR address #273

### DIFF
--- a/python/sourced/engine/engine.py
+++ b/python/sourced/engine/engine.py
@@ -108,6 +108,50 @@ class Engine(object):
         """
         return self.__implicits.parseUASTNode(data)
 
+    def from_metadata(self, db_path, db_name='engine_metadata.db'):
+        """
+        Registers in the current session the views of the MetadataSource so the
+        data is obtained from the metadata database instead of reading the
+        repositories with the DefaultSource.
+
+        :param db_path: path to the folder that contains the database.
+        :type db_path: str
+        :param db_name: name of the database file (engine_metadata.db) by default.
+        :type db_name: str
+        :returns: the same instance of the engine
+        :rtype: Engine
+        """
+
+        self.__engine.fromMetadata(db_path, db_name)
+        return self
+
+    def from_repositories(self):
+        """
+        Registers in the current session the views of the DefaultSource so the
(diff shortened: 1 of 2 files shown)


python/sourced/engine/engine.py
```diff
@@ -108,6 +108,50 @@ def parse_uast_node(self, data):
...
+
+        :returns: the same instance of the engine
+        :rtype: Engine
+        """
+
+        return self.__engine.fromRepositories()
+
+    def save_metadata(self, path, db_name='engine_metadata.db'):
+        """
+        Saves all the metadata in a SQLite database on the given path as "engine_metadata.db".
```
erizocosmico: s/as "engine_metadata.db"/with the given filename (which if not given is "engine_metadata.db"/

python/sourced/engine/engine.py
```diff
@@ -108,6 +108,50 @@ def parse_uast_node(self, data):
...
+        If the database already exists, it will be overwritten. The given path must exist and
+        must be a directory, otherwise it will throw a [[SparkException]].
+        Saved tables are repositories, references, commits and tree_entries. Blobs are not saved.
+
+        :param path:   where database with the metadata will be stored.
+        :param db_name: name of the database file
```
erizocosmico: add that `(default is engine_metadata.db)`

python/test/test_engine.py
```diff
@@ -217,3 +217,15 @@ def test_extract_tokens(self):
...
+
+    def test_metadata(self):
+        tmpdir = tempfile.mkdtemp()
+
+        self.engine.save_metadata(tmpdir)
+        db_path = path.join(tmpdir, 'engine_metadata.db')
+        self.assertTrue(path.exists(db_path))
+
+        engine = self.engine.from_metadata(tmpdir)
+        self.assertEqual(self.engine.session, engine.session)
```
erizocosmico: This check is kinda pointless, the session instance is never changed. Instead, what do you think about checking that a dataframe has the same output in both datasources?

For example:

```python
expected = self.engine.repositories.count()
obtained = self.engine.from_metadata(tmpdir).repositories.count()
self.assertEqual(obtained, expected)
```

PR #288, opened 2018-01-04
add is_remote column to references table

erizocosmico:
Adds `is_remote` column to `references` table so an user can identify if a reference was remote or not even after the rewrite from `refs/remotes/$REMOTE_NAME/$BRANCH` to `refs/heads/$BRANCH`.

### DIFF
--- a/src/main/scala/tech/sourced/engine/Engine.scala
+++ b/src/main/scala/tech/sourced/engine/Engine.scala
@@ -254,8 +254,6 @@ class Engine(session: SparkSession,
       dbFile.toFile.delete()
     }
 
-    implicit val session: SparkSession = this.session
-
     val properties = new Properties()
     properties.put("driver", "org.sqlite.JDBC")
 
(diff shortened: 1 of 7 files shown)


src/main/scala/tech/sourced/engine/Engine.scala
```diff
@@ -317,3 +315,35 @@ object Engine {
...
+
+    /**
+      * Returns a new DataFrame with the given string array column converted to
+      * a column with the content of the array joined by "|".
+      *
+      * @param column column name
+      * @return new dataframe
+      */
+    def withStringArrayColumnAsString(column: String): DataFrame =
+      df.withColumn(column, ConcatArrayUDF(df(column), lit("|")))
```
mcarmonaa: You need to pass the session as an implicit parameter to the UDF

erizocosmico:
Fixed @mcarmonaa

mcarmonaa:
@erizocosmico Is still needed the `ConcatArrayUDF`object? it seems that `withStringArrayColumnAsString` does the same thing

erizocosmico:
@mcarmonaa `withStringArrayColumnAsString` is the method that replaces the column, the other thing is an UDF because you cannot transform this without an UDF.

mcarmonaa:
Oh forget it, I was looking at the code in master and I got mixed up

PR #297, opened 2018-01-05
docs on raw repos usage, getRemoteReferences added

erizocosmico:


erizocosmico:
tests fixed

erizocosmico:
@mcarmonaa 👍, I'll add that raw means standard and bare, but I don't know if it's useful to add another example for bare, because the code is exactly the same except for `s/standard/bare/`

erizocosmico:
done @mcarmonaa

jfontan:
I believe it's a bit confusing. You specify "standard" as parameter that means "raw" that include both "standard" and "bare" repos. Is this correct?

erizocosmico:
Nope, raw is a term to refer to both standard and bare (which is not valid as a parameter). Only valid parameters are `standard`, `bare` and `siva`.

For demo purposes, I had to pick one of them and it was `standard`, because both work in the exact same way.

jfontan:
Got it, thanks for the clarification. Somehow I understood that only "standard" and "siva" were allowed as parameter.

Issue #286, opened 2018-01-04
[Documentation/Examples] Raw repository reading


eiso:
An example for reading from raw repositories should be included in the example notebooks and in the readme.

eiso:
The current example notebook fails:

```from sourced.engine import Engine
from pyspark.sql import SparkSession
from pyspark.sql.functions import *

spark = SparkSession.builder\
.master("local[*]").appName("Examples")\
.getOrCreate()

engine = Engine(spark, "/repositories")```

```---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-49cf0ae5a267> in <module>()
      5 spark = SparkSession.builder.master("local[*]").appName("Examples").getOrCreate()
      6 
----> 7 engine = Engine(spark, "/repositories")

TypeError: __init__() missing 1 required positional argument: 'repos_format'
```

erizocosmico:
Done in #297

PR #298, opened 2018-01-08
Modified the example notebook

eiso:
Adding some more examples and adding sampling and limits to give users faster results the first time they run it. I also added  some more printSchema() statements so that the first time user unknown to Spark can get an understanding on how it works. It also now passes on the param to read from raw repositories, which is the more common behaviour for most users.

This is notebook IMO still has too few examples and I'd like to encourage everyone to add more interesting ones.

### DIFF
--- a/examples/notebooks/Example.ipynb
+++ b/examples/notebooks/Example.ipynb
@@ -1,11 +1,16 @@
 {
  "cells": [
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Welcome to source{d} engine"
+   ]
+  },
+  {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": true
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "from sourced.engine import Engine\n",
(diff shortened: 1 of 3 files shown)


eiso:
One thing, that is completely open for discussion, is if to include the sample results I added.

erizocosmico:
You need to rebase

examples/notebooks/Example.ipynb
```diff
@@ -16,7 +14,76 @@
     ".master(\"local[*]\").appName(\"Examples\")\\\n",
     ".getOrCreate()\n",
     "\n",
-    "engine = Engine(spark, \"/repositories\")"
+    "engine = Engine(spark, \"/repositories\", \"standard\")"
```
erizocosmico: why is the example using `standard` instead of siva? Isn't the default thing when you run the jupyter siva files?

examples/notebooks/Example.ipynb
```diff
@@ -16,7 +14,76 @@
...
+   "cell_type": "code",
+   "execution_count": 119,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "root\n",
+      " |-- id: string (nullable = false)\n",
```
erizocosmico: it's a bit tricky to have output here, specially the schema, because it may get outdated quickly if we add/remove changes

examples/notebooks/Example.ipynb
```diff
@@ -16,7 +14,76 @@
...
+   "metadata": {},
+   "source": [
+    "### Show all repositories in /repositories"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Even though there are 2 repositories, 4 are shown because of the remote and local references in git"
```
erizocosmico: isn't this too ad-hoc for a specific scenario mounted to the jupyter to be in the example?

eiso:
@miguel all your suggestions are valid and I'll change accordingly

@campoy just run

examples/notebooks/Example.ipynb
eiso: +1

examples/notebooks/Example.ipynb
eiso: +1

examples/notebooks/Example.ipynb
eiso: For any user who docker pulls this repo, he'll be using non siva.

As @mcuadros mentioned it to me, sourced will be one of the few companies who actually use siva files because we have such large scale implementations. Most individual devs will go for standard repos. This might make this notebook conflict though with the hosted version. @mcuadros please advise.

eiso:
@erizocosmico I think all of your comments are addressed, I have added more examples, so a review is welcome (sorry for the pain in the ass, it is do it with iPython notebooks)

examples/notebooks/Example.ipynb
erizocosmico: Then we should change the dockerfile and so on to clone some example repos, because AFAIK, running the default container puts siva files in `/repositories`. The "didn't touch anything, just ran it" case should work, which is not the case right now because siva files are mounted. So, if we want that case to be for standard repositories, we should change the mounted files accordingly.

eiso:
I agree with you but I didn't find any default siva files currently in the engine-jupyter image.

@mcuadros do you agree with the change of having raw as default. It's more natural, and while the container runs you can easily add more repos by cloning and reinitializing the engine.

campoy:
Not sure where to put this feedback, but I'm still unable to run this on my laptop

```bash
$ docker volume create bblfsh
$ docker run -p 8888:80 --link bblfshd bblfsh/dashboard -bblfsh-addr bblfshd:9432
$ docker exec -it bblfshd bblfshctl driver install --all
$ docker run --name engine-jupyter --rm -it -p 8080:8080 -v $(pwd)/repos:/repositories --link bblfshd:bblfshd srcd/engine-jupyter
```

When I run the first cell I get this error.

screen shot 2018-01-12 at 4 26 58 pm

Not sure how to fix it, to be honest

eiso:
It's this pending PR that fixes that. Git clone this PR and upload the example notebook that's here instead. It was because of a breaking API change on needing to add the type of repository you're reading, see https://github.com/src-d/engine/issues/286

examples/notebooks/Example.ipynb
```diff
@@ -16,50 +21,146 @@
     ".master(\"local[*]\").appName(\"Examples\")\\\n",
     ".getOrCreate()\n",
     "\n",
-    "engine = Engine(spark, \"/repositories\")"
+    "engine = Engine(spark, \"/repositories\", \"standard\")"
    ]
```
campoy: I had some issue with this because I had not moved to `standard`. The problem is that it would not fail here, instead it was failing the line after.

Also when it worked I did not realize it had actually finished!

Let's add this line at the end of the cell to fix both problems:

```python
print("%d repositories successfully loaded" % engine.repositories.count())
```

examples/notebooks/Example.ipynb
```diff
@@ -16,50 +21,146 @@
...
    "source": [
-    "### Get all the blobs of all head commits"
+    "### Show all repositories in /repositories"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Double the number of repositories maybe shown because of remote and local references in git"
```
campoy: Maybe? I saw this, but this line seems pretty ambiguous.

erizocosmico:
Ping @eiso?

eiso:
Thanks for the ping @erizocosmico, I somehow missed Francesc's review. I'll make the change today and I believe it's ready to merge then. Going for standard repositories as the default, makes most sense for new users, less for them to learn. Once they to to scale, they can use borges and siva.

eiso:
@erizocosmico I made the final changes. 

@campoy could you decide which standard repos you want in the docker container, and paste the list here or modify the docker file directly

Once @campoy has listed them, it would be great to make a tag release tomorrow because of: https://github.com/src-d/minutes/pull/116/files#diff-ddd24239d7d2e1f32dd0b3dc5b363fa8R21

erizocosmico:
@eiso there are conflicts

PR #300, opened 2018-01-10
better reporting of enry exceptions

erizocosmico:
That's the most we can do at the UDF level for the data we have, print a log with the failure and the filename. And only considering this throws an exception, if it simply makes everything die or it does not this is for nothing.

### DIFF
--- a/src/main/scala/tech/sourced/engine/udf/ClassifyLanguagesUDF.scala
+++ b/src/main/scala/tech/sourced/engine/udf/ClassifyLanguagesUDF.scala
@@ -1,12 +1,13 @@
 package tech.sourced.engine.udf
 
+import org.apache.spark.internal.Logging
 import org.apache.spark.sql.expressions.UserDefinedFunction
 import org.apache.spark.sql.functions.udf
 import org.apache.spark.sql.{Column, SparkSession}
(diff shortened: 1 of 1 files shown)


src/main/scala/tech/sourced/engine/udf/ClassifyLanguagesUDF.scala
```diff
@@ -29,7 +30,13 @@ case object ClassifyLanguagesUDF extends CustomUDF {
     if (isBinary) {
       None
     } else {
-      val lang = Enry.getLanguage(path, content)
+      val lang = try {
+        Enry.getLanguage(path, content)
+      } catch {
+        case e: Exception =>
```
ajnavarro: RuntimeExceptions will not be catched here. You should change this for ` case e @ (_ : RuntimeException | _ : Exception) => ....`

bzz: So true!.. Very good to be reminded of this in Scala

erizocosmico: fixed

PR #306, opened 2018-01-15
get only the last commit of a reference by default

### LINKED ISSUE #290: Make getFirstReferenceCommit the default behavior
## Expected Behavior

Instead of retrieving all commits of all references of all repositories, only the first reference commit (aka the current state of the reference) should be retrieved by default.
Instead of a `getFirstReferenceCommit` we can add a `getAllReferenceCommits` for the old behavior.

## Current Behavior

Retrieves all commits from all references from all repos by default.

## Possible Solution

Make `getFirstReferenceCommit` default, add a `getAllReferenceCommits` to opt-in into old default behavior.

## Steps to Reproduce (for bugs)

n/a

## Context

The rationale behind this is that most of the time what you want is `getFirstReferenceCommit`, which is the obvious thing a person would expect engine to do. If you want the more detailed output provided by `getAllReferenceCommits` you are opting in into this, which you already know will severely harm the performance of the job you're running.

So, this as a default that makes more sense for the following reasons:

- More obvious behavior for newcomers
- Default that does not severely impact performance
- Previous effect can still be achieved

## Your Environment

n/a

Also, /cc @vmarkovtsev @bzz as users what do you think about this?

### LINKED ISSUE #303: Or filters are not correctly compiled
## Expected Behavior

Or filters should be compiled into or filters

## Current Behavior

Or filters are compiled into And filters in `matchingCases` because they extend from `BinaryFilter` and what it does is `left.matchingCases ++ right.matchingCases`

## Possible Solution

Support only and filters for the time being until a refactor of the filters is properly done. It's not like it's not gonna work, Spark will take care of it. It's just that it'd be more efficient if we did it.

https://github.com/src-d/engine/blob/master/src/main/scala/tech/sourced/engine/util/Filters.scala#L80 and `s/BinaryFilter/AndFilter/` in that line would do the trick.

## Steps to Reproduce (for bugs)

Create any or filter

## Context

I came up with this while currently working on the implementation of the ENIP 004. It's proven to be trickier than I initially thought. So far, with the current limitations our filter system has I see it very unlikely that we're going to be able to provide an `or` filter solution for the time being.

## Your Environment (for bugs)

happens in all envs

erizocosmico:
This implements the ENIP-004, which makes the default behavior returning only
the last commit in a reference unless specifically asked to retrieve them all.
It has sort of a hack for making it work with the metadata source, since it
does not make use of the CommitIterator, which has a less hacky implementation.

Also, there's a complete rewrite of the filters and how they are used in the
iterators to allow more operators than just equality. As a side effect, 'OR'
operator is no longer supported. It was supported before, but only because it
was converted to an AND (and thus being a bug).

Fixes #303.
Closes #290.

### Pending tasks

- [x] update documentation
- [x] update examples
- [x] update examples inside the python package

### Warning!

This contains breaking changes and should be release in version 0.4.x! Be careful when merging this.

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -84,7 +84,7 @@
  </check>
  <check level="error" class="org.scalastyle.scalariform.MethodLengthChecker" enabled="true">
   <parameters>
-   <parameter name="maxLength"><![CDATA[80]]></parameter>
+   <parameter name="maxLength"><![CDATA[150]]></parameter>
   </parameters>
  </check>
  <check level="error" class="org.scalastyle.scalariform.MethodNamesChecker" enabled="true">
--- a/src/main/scala/tech/sourced/engine/QueryBuilder.scala
+++ b/src/main/scala/tech/sourced/engine/QueryBuilder.scala
@@ -188,7 +188,8 @@ private[engine] object QueryBuilder {
         // If we can make sure compileFilter supports all filters, we can remove this check.
         val or = Seq(f1, f2).flatMap(compileFilter)
         if (or.size == 2) {
-          or.map(p => s"($p)").mkString(" OR ")
+          val clause = or.map(p => s"($p)").mkString(" OR ")
+          s"($clause)"
         } else {
           null
         }
--- a/src/main/scala/tech/sourced/engine/iterator/CommitIterator.scala
+++ b/src/main/scala/tech/sourced/engine/iterator/CommitIterator.scala
@@ -6,7 +6,7 @@ import org.eclipse.jgit.api.Git
 import org.eclipse.jgit.errors.IncorrectObjectTypeException
 import org.eclipse.jgit.lib.{ObjectId, Ref, Repository}
 import org.eclipse.jgit.revwalk.RevCommit
-import tech.sourced.engine.util.{CompiledFilter, Filter}
+import tech.sourced.engine.util.{CompiledFilter, Filters}
 
 import scala.collection.JavaConverters.iterableAsScalaIterableConverter
 
@@ -24,15 +24,15 @@ class CommitIterator(finalColumns: Array[String],
                      filters: Seq[CompiledFilter])
   extends ChainableIterator[ReferenceWithCommit](finalColumns, prevIter, filters) {
 
-  /** @inheritdoc */
+  /** @inheritdoc*/
   override protected def loadIterator(filters: Seq[CompiledFilter]): Iterator[ReferenceWithCommit] =
     CommitIterator.loadIterator(
       repo,
       Option(prevIter).map(_.currentRow),
-      filters.flatMap(_.matchingCases)
+      Filters(filters)
     )
 
-  /** @inheritdoc*/
+  /** @inheritdoc */
   override protected def mapColumns(obj: ReferenceWithCommit): RawRow = {
     val (repoId, refName) = RootedRepo.parseRef(repo, obj.ref.getName)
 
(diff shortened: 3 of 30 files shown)


erizocosmico:
CI passing now.

scalastyle-config.xml
```diff
@@ -84,7 +84,7 @@
  </check>
  <check level="error" class="org.scalastyle.scalariform.MethodLengthChecker" enabled="true">
   <parameters>
-   <parameter name="maxLength"><![CDATA[80]]></parameter>
+   <parameter name="maxLength"><![CDATA[200]]></parameter>
```
ajnavarro: Instead of change this file, you can use special comments for scala style to do not check specific parts of the code (If it's is absolutely needed): http://www.scalastyle.org/configuration.html#comment-filters

src/main/scala/tech/sourced/engine/QueryBuilder.scala
```diff
@@ -188,7 +188,7 @@ private[engine] object QueryBuilder {
         // If we can make sure compileFilter supports all filters, we can remove this check.
         val or = Seq(f1, f2).flatMap(compileFilter)
         if (or.size == 2) {
-          or.map(p => s"($p)").mkString(" OR ")
+          s"(${or.map(p => s"($p)").mkString(" OR ")})"
```
ajnavarro: This is really difficult to read. I would prefer to separate this in a variable assignment and then print it.

src/main/scala/tech/sourced/engine/iterator/CommitIterator.scala
```diff
@@ -148,12 +141,16 @@ class RefWithCommitIterator(repo: Repository,
           case _: IncorrectObjectTypeException => null
           // TODO: This reference is pointing to a non commit object, log this
         }
+
+      if (justFirst && commits != null) {
```
ajnavarro: Maybe can be interesting for future improvements send the max number of commits to fetch instead of a boolean. What do you think?

scalastyle-config.xml
erizocosmico: Yep, but it's a bit dirty and the max length check is kinda pointless

erizocosmico:
Done @ajnavarro

scalastyle-config.xml
ajnavarro: I aggree that 80 is really small. Should we leave this with the dafault value? 150? http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/sizes/MethodLengthCheck.html

I think is enough and readable for most of the cases.

src/main/scala/tech/sourced/engine/QueryBuilder.scala
```diff
@@ -188,7 +188,8 @@ private[engine] object QueryBuilder {
         // If we can make sure compileFilter supports all filters, we can remove this check.
         val or = Seq(f1, f2).flatMap(compileFilter)
         if (or.size == 2) {
-          or.map(p => s"($p)").mkString(" OR ")
+          val clause = or.map(p => s"($p)").mkString(" OR ")
+          s"($clause)"
```
mcarmonaa: Isn't clause already a string?

erizocosmico: Yes, note the `(` and `)` enclosing it. It's enclosing the clause in parenthesis

scalastyle-config.xml
erizocosmico: Sounds fine

src/main/scala/tech/sourced/engine/QueryBuilder.scala
mcarmonaa: Oh, I see :+1:

erizocosmico:
@ajnavarro changed the config to 150

erizocosmico:
Before merging, there should be an update to the docs and examples. Let's not merge this yet. Also, there's a few PR's I'd rather merge before this one, which has breaking changes.

erizocosmico:
done, docs updated

PR #307, opened 2018-01-15
upgrade siva-java so it handles files bigger than 2GB

erizocosmico:
(CI might fail as the version is still not available in maven)


erizocosmico:
CI passing, now

PR #313, opened 2018-01-18
Make Engine's session member public

mcarmonaa:
I think is valuable to have the `session` member publicly available since it allows you to pass an `Engine` object throw your code containing all you need from `spark`.

For example, when engine is used by a third app you could want import `spark.sqlContext.implicits._` in some parts of your code. Having the session member available you don't need to pass the spark session and engine to those parts, just engine and use `import engine.session.sqlContext.implicits._`


erizocosmico:
Makes sense to me

PR #317, opened 2018-01-19
allow bucketized raw repositories, limit siva files

### LINKED ISSUE #235: Allow standard repositories to be bucketized
If we want to read repositories from a folder called `/repos` the repositories should be at the next level (`/repos/foo` `/repos/bar`). We want to be able to read repositories with more complex paths, or bucketized (`/repos/f/foo` `/repos/b/bar` `/repos/b/baz`)

### LINKED ISSUE #316: [bug] Handle *.siva.copy files
## Expected Behavior
If there are *.siva.copy files, engine does not fail the job.

## Current Behavior
If there are *.siva.copy files, engine kills the job with

```
Caused by: tech.sourced.siva.SivaException: Exception at file 0b3678c14ab5af5cd5019b482405b5bdc0f17d26.siva.copy: At Index footer, index size: Java implementation of siva doesn't support values greater than 9223372036854775807
```

## Possible Solution

1. Skip offending files
2. Fix Java to handle such files

## Context
I am running queries on 150k repositories and want them to finish in any case.

Example: `hdfs://hdfs-namenode/apps/borges/siva-copy`

erizocosmico:
Fixes #316 
Closes #235 

This allows bucketization of raw repositories and limits the files of the `siva` repo format to `*.siva` files.

### DIFF
--- a/src/main/scala/tech/sourced/engine/provider/RepositoryRDDProvider.scala
+++ b/src/main/scala/tech/sourced/engine/provider/RepositoryRDDProvider.scala
@@ -62,43 +62,45 @@ object RepositoryRDDProvider {
 
   /**
     * Generates an RDD of [[RepositorySource]] with the repositories at the given path.
-    * Allows bucketting of siva files, but not bucketting of repositories or bare repoitories.
+    * Allows bucketing of siva files and raw repositories.
     *
     * @param sc                 Spark Context
     * @param path               path to get the repositories from
     * @param repositoriesFormat format of the repositories inside the provided path
     * @return generated RDD
     */
-  private def generateRDD(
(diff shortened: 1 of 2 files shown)


eiso:
Does this mean it will identify all git repos across an entire home folder for instance?

erizocosmico:
@eiso not exactly. It's not recursive, it only looks for repositories with a maximum depth of 1.

if you have:

```
user
|-- repo1
|-- repo2
|-- repos
   |-- repo3
|-- repos-2
   |-- repos-3
       |-- repo 4
```

only `repo1`, `repo2`, and `repo3` would be found, but not `repo4`

PR #326, opened 2018-01-31
Limit the number of repositories kept in the pool

jfontan:
The limit is set to the double of the number of parallelism. Without thischange there could be OutOfMemory problems when processing big repos.

Fixes #319

PR #327, opened 2018-01-31
[WIP] Limit the number of repositories kept in the pool

jfontan:
The limit is set to the double of the number of parallelism. Without this change there could be `OutOfMemory` problems when processing big repos.


jfontan:
Fixed the previous PR #326

Issue #339, opened 2018-03-06
java.io.InvalidClassException: tech.sourced.engine.util.EqualFilter; no valid constructor


zurk:
## Expected Behavior
Engine should be able to analyze PGA and extract UAST from source code.

## Current Behavior
It fails with exception on worker:
```
18/03/06 14:20:53 ERROR Utils: Exception encountered
java.io.InvalidClassException: tech.sourced.engine.util.EqualFilter; no valid constructor
	at java.io.ObjectStreamClass$ExceptionInfo.newInvalidClassException(ObjectStreamClass.java:150)
	at java.io.ObjectStreamClass.checkDeserialize(ObjectStreamClass.java:790)
	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1782)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1353)
	at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2018)
	at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:1942)
	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1808)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1353)
	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:373)
	at scala.collection.immutable.List$SerializationProxy.readObject(List.scala:479)
	at sun.reflect.GeneratedMethodAccessor34.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at java.io.ObjectStreamClass.invokeReadObject(ObjectStreamClass.java:1058)
	at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:1909)
	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1808)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1353)
	at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2018)
	at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:1942)
	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1808)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1353)
[...]

PR #341, opened 2018-03-14
Add default constructors to all Filters

bzz:
This PR follows the advice in [SI-4278](https://issues.scala-lang.org/browse/SI-4278) and makes filters serializable, thus avoids cluster-mode runtime exceptions like https://github.com/src-d/engine/issues/339

Please, feel free to provide a feedback if there is a better way of doing it.


r0mainK:
I tested this out on a locally deployed spark standalone, as well as on the gpu-cluster, it got rid of all the problems we were experiencing

PR #360, opened 2018-04-06
Add detailed debug info when an iterator crashes

carlosms:
This PR is a proposal to add extra information when an untreated exception is caught in `CleanupIterator`.

The idea is that iterators can extend a new `CrashReporter` trait, and the global try/catch in `CleanupIterator` will check if the current iterator can add details to the exception.

Note: Each iterator will probably have extra useful debug details (maybe in `currentRow`?) that can be added later. I wanted to make this PR as simple as possible to request feedback.

Now the exceptions will look like this:
```
scala> engine.getRepositories.getReferences.getCommits.count()
[Stage 0:>                                                          (0 + 3) / 3]18/04/06 20:24:08 ERROR Executor: Exception in task 2.0 in stage 0.0 (TID 2)
java.lang.Exception: RefWithCommitIterator crashed for Repository[/tmp/spark-2465f2ae-b23a-429b-9d78-1529ccbd3852/processing-repositories/3605C2257CBB88D2A0F833494C945F93/13ffbec95ff31d0b0a2fb50e2e63145430a152da.siva]; urls https://github.com/angular/protractor, https://github.com/juliemr/protractor, https://github.com/juliemr/protractor, https://github.com/angular/protractor
	at tech.sourced.engine.iterator.CommitIterator.exceptionDetails(CommitIterator.scala:68)
	at tech.sourced.engine.iterator.CleanupIterator.hasNext(CleanupIterator.scala:31)
	at scala.collection.Iterator$$anon$12.hasNext(Iterator.scala:438)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIterator.agg_doAggregateWithoutKey$(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIterator.processNext(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenExec$$anonfun$8$$anon$1.hasNext(WholeStageCodegenExec.scala:395)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
	at org.apache.spark.shuffle.sort.BypassMergeSortShuffleWriter.write(BypassMergeSortShuffleWriter.java:125)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:96)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:53)
	at org.apache.spark.scheduler.Task.run(Task.scala:108)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:335)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.Exception: Forced exception for testing
	at tech.sourced.engine.iterator.RefWithCommitIterator.hasNext(CommitIterator.scala:147)
	at tech.sourced.engine.iterator.ChainableIterator.hasNext(ChainableIterator.scala:89)
	at org.apache.spark.InterruptibleIterator.hasNext(InterruptibleIterator.scala:37)
	at tech.sourced.engine.iterator.CleanupIterator.hasNext(CleanupIterator.scala:23)
	... 15 more
```

### DIFF
--- a/src/main/scala/tech/sourced/engine/iterator/CleanupIterator.scala
+++ b/src/main/scala/tech/sourced/engine/iterator/CleanupIterator.scala
@@ -11,7 +11,7 @@ import org.apache.spark.{InterruptibleIterator, TaskContext}
   * @tparam T type of the rows in the iterator
   */
 class CleanupIterator[T](it: Iterator[T], cleanup: => Unit)
-  extends InterruptibleIterator[T](TaskContext.get(), it) {
+    extends InterruptibleIterator[T](TaskContext.get(), it) {
 
   /** @inheritdoc
     *
@@ -26,7 +26,7 @@ class CleanupIterator[T](it: Iterator[T], cleanup: => Unit)
       }
       hasNext
     } catch {
-      case e @ (_: Exception | _: RuntimeException) =>
+      case e: Exception =>
         it match {
           case it: ChainableIterator[_] =>
             throw new RepositoryException(it.repo, e)
--- a/src/main/scala/tech/sourced/engine/iterator/CleanupIterator.scala
+++ b/src/main/scala/tech/sourced/engine/iterator/CleanupIterator.scala
@@ -26,17 +26,22 @@ class CleanupIterator[T](it: Iterator[T], cleanup: => Unit)
       }
       hasNext
     } catch {
-      case e: Exception =>
-        it match {
-          case it: ChainableIterator[_] =>
-            throw new RepositoryException(it.repo, e)
-          case _ =>
-            throw e
-        }
       case e: Throwable =>
-        throw e
-    } finally {
-      val _ = cleanup
+        val detailedException = (
+          e match {
+            case e: Exception =>
+              it match {
+                case it: ChainableIterator[_] =>
+                  new RepositoryException(it.repo, e)
+                case _ =>
+                  e
+              }
+            case _ =>
+              e
+          })
+
+        val _ = cleanup
+        throw detailedException
     }
   }
 
--- a/src/main/scala/tech/sourced/engine/iterator/CleanupIterator.scala
+++ b/src/main/scala/tech/sourced/engine/iterator/CleanupIterator.scala
@@ -26,9 +26,17 @@ class CleanupIterator[T](it: Iterator[T], cleanup: => Unit)
       }
       hasNext
     } catch {
+      case e @ (_: Exception | _: RuntimeException) =>
+        it match {
+          case it: ChainableIterator[_] =>
+            throw new RepositoryException(it.repo, e)
+          case _ =>
+            throw e
+        }
       case e: Throwable =>
-        val _ = cleanup
         throw e
+    } finally {
+      val _ = cleanup
     }
   }
 
--- a/src/main/scala/tech/sourced/engine/iterator/RepositoryException.scala
+++ b/src/main/scala/tech/sourced/engine/iterator/RepositoryException.scala
@@ -1,0 +1,38 @@
+package tech.sourced.engine.iterator
+
+import org.eclipse.jgit.lib.Repository
+import scala.collection.JavaConverters.iterableAsScalaIterableConverter
+
+/**
+  * Exception to add repository debug information to any
+  * uncontrolled exception. It does not add a stacktrace level.
+  *
+  * @param repo Repository that was beeing iterated
+  * @param cause Original exception
+  */
+class RepositoryException(repo: Repository, cause: Throwable)
+    extends Exception(
+      s"Repository error with data: ${RepositoryException.repoInfo(repo)}",
+      cause,
+      true,
+      false) {}
+
+object RepositoryException {
+
+  /**
+    * Returns a string with a debug description of the repository
+    * @param repo Repository to describe
+    * @return
+    */
+  def repoInfo(repo: Repository): String = {
+    val c = repo.getConfig
+    val remotes = c.getSubsections("remote").asScala
+    val urls = remotes.flatMap(r => c.getStringList("remote", r, "url"))
+
+    if (urls.isEmpty) {
+      s"${repo.toString}"
+    } else {
+      s"${repo.toString}; urls ${urls.mkString(", ")}"
+    }
+  }
+}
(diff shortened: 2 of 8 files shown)


bzz:
Nice idea! 👏 

One suggestions - instead of throwing generic `Exception` and adding information from `repoInfo(repo)` to it's message in String, do you think it would be worth having a separate application-specific class? 

I.e a dedicated `GitRepoIterationException` with a field for details about the origin of the data/git repo causing troubles that can be used to generate appropriate error message by this class itself might improve maintainability of the solution.

src/main/scala/tech/sourced/engine/iterator/CrashReporter.scala
```diff
@@ -0,0 +1,26 @@
...
+    */
+  def exceptionDetails(e: Throwable): Throwable
+
+  /**
+    * Returns a string with a debug description of the repository
+    * @param repo Repository to describe
+    * @return
+    */
+  protected def repoInfo(repo: Repository): String = {
+    val c = repo.getConfig
```
ajnavarro: This code should be totally error-safe

carlosms:
Branch rebased.
@bzz, @ajnavarro, thanks for the suggestion, I've created a new RepositoryException. A nice thing about it is that it can avoid adding one more level to the stacktrace, so the new exception is just the new message followed by the original stacktrace.

@ajnavarro can you be more specific about `repo.getConfig` not being safe? Can it return a null? I think the code below it should be fine, `remotes` and `urls` are Iterables and the code should be fine if these are empty, right? This code is now part of RepositoryException.

src/main/scala/tech/sourced/engine/iterator/RepositoryException.scala
```diff
@@ -0,0 +1,40 @@
...
+
+object RepositoryException {
+
+  /**
+    * Returns a string with a debug description of the repository
+    * @param repo Repository to describe
+    * @return
+    */
+  def repoInfo(repo: Repository): String = {
+    val c = repo.getConfig
```
bzz: My guess is that by https://github.com/src-d/engine/pull/360#discussion_r180057378 @ajnavarro means handling in code cases like this https://github.com/eclipse/jgit/blob/master/org.eclipse.jgit/src/org/eclipse/jgit/internal/storage/file/FileRepository.java#L402 where particular jgit methods i.e `.getConfig()` here, could throw something.

carlosms: Oh nice, runtime exception without any hint in the signature!
I guess the whole `repoInfo` method could have a main try/catch to account for any exception, and this way we make sure the method finishes and `CleanupIterator` will be guaranteed to call `cleanup`.

ajnavarro:
@carlosms Can you rebase and remove the duplicated method introduced here please? https://github.com/src-d/engine/pull/360

src/main/scala/tech/sourced/engine/iterator/RepositoryException.scala
ajnavarro: > I guess the whole repoInfo method could have a main try/catch to account for any exception, and this way we make sure the method finishes and CleanupIterator will be guaranteed to call cleanup.

carlosms:
Thanks @ajnavarro for sitting with me and proposing a better approach. Now `ChainableIterator` contains a repository so it can be used to a `RepositoryException` in `CleanupIterator`.

The branch is now rebased and squashed.

src/main/scala/tech/sourced/engine/iterator/CleanupIterator.scala
```diff
@@ -26,9 +26,17 @@ class CleanupIterator[T](it: Iterator[T], cleanup: => Unit)
       }
       hasNext
     } catch {
+      case e @ (_: Exception | _: RuntimeException) =>
```
ajnavarro: Sorry, but I forgot that just doing `case e:Exception` it will work in the same way, because RuntimeException extends Exception.

src/main/scala/tech/sourced/engine/iterator/RepositoryException.scala
carlosms: done, thank you both.

src/main/scala/tech/sourced/engine/iterator/CleanupIterator.scala
carlosms: fixed, thank you

bzz:
👍  

May be it's also worth looking up and posting here a reason for CI failure?

src/main/scala/tech/sourced/engine/iterator/RepositoryException.scala
```diff
@@ -0,0 +1,43 @@
...
+      val urls = remotes.flatMap(r => c.getStringList("remote", r, "url"))
+
+      if (urls.isEmpty) {
+        s"${repo.toString}"
+      } else {
+        s"${repo.toString}; urls ${urls.mkString(", ")}"
+      }
+    } catch {
+      case e: Throwable =>
+        s"Exception in RepositoryException.repoInfo ${e.getMessage}"
```
bzz: Is there chance to add something like
```scala
val sivaFile = try {
  repo.toString
} catch {
  case _ : Throwable "Unknown .siva file"
}
```

at the beginning and then include `sivaFile` in this message?  This would allow us to keep error reporting quality as high as possible, given that exceptions most probably are not going to happen in `toString` method.

ajnavarro: Engine can handle repositories in different ways, so maybe is better to use `Unknown repository path` instead of `Unknown .siva file`

carlosms:
@bzz I'm looking into it, will report here if I find something useful

src/main/scala/tech/sourced/engine/iterator/RepositoryException.scala
carlosms: Nice suggestion, done.

carlosms:
It took me a while. We had put a cleanup call in the finally block, which is fine after the catch. But that was also doing a cleanup when there was no exception, so the tests were failing complaining about missing resources.

It should be fixed now.

ajnavarro:
Thanks a lot @carlosms

PR #369, opened 2018-04-10
ci: upload images in master as well

### LINKED ISSUE #366: Push Docker images for development versions (master)
## Expected Behavior

`dev-<commit>` tags for engine-jupyter should be available at Docker Hub.

## Current Behavior
They are not pushed.

## Possible Solution
Use functionality already available at src-d/ci:
https://github.com/src-d/ci/blob/beb8af9aa6417e701751858d449bbdfe6f3ee3bd/Makefile.main#L171

## Context
This will make it easier to test latest development version on our k8s clusters.

erizocosmico:
Fixes #366


ajnavarro:
`latest` tag is set right now when a new container is pushed, but I cannot see where, because we are not using `DOCKER_PUSH_LATEST` environment variable to do it. On this case, `latest` version must not point to a `dev` version.

PR #370, opened 2018-04-10
iterator: traverse all commits when iterator is not chained

### LINKED ISSUE #368: commits table with filter index does not walk the full history
## Expected Behavior

This query should output something higher than 0 if there is any repository with more than one commit in the same reference:

```sql
SELECT COUNT(*) FROM commits WHERE index > 0
```

## Current Behavior
This works for me with local Spark + engine 0.5.5 (both `spark-shell --package` and `engine-jupyter`, both local filesystem and HDFS).

But in cluster mode (pipeline-staging cluster), it does not work:

```
scala> sc.master
res34: String = spark://[master host]:7077

scala> sc.version
res35: String = 2.2.0

scala> import tech.sourced.engine._
import tech.sourced.engine._

scala> val engine = Engine(spark, "hdfs://[hdfs namenode host]/100-siva", "siva")
engine: tech.sourced.engine.Engine = tech.sourced.engine.Engine@10db002c

scala> spark.registerUDFs

scala> print(s"${engine.getRepositories.count()} repositories successfully loaded")
149 repositories successfully loaded                             

scala> spark.sql("SELECT COUNT(*) FROM commits").show(1, false)
+--------+                                                                      
|count(1)|
+--------+
|100     |
+--------+

scala> spark.sql("SELECT COUNT(*) FROM commits WHERE index > 0").show(1, false)
+--------+                                                                      
|count(1)|
+--------+
|0       |
+--------+
```

## Possible Solution
No idea.

## Steps to Reproduce
I have not tried with other clusters so far, so I'm not sure if this fails with cluster mode in general, or with our Spark cluster in staging (Spark 2.2.0).

## Context
Trying to executing some queries with engine-jupyter + staging cluster.

## Your Environment
* Version used: Spark 2.2.0, engine 0.5.5
* Operating System and version: pipeline-staging cluster on kubernetes
* Some needed resources to reproduce the problem: I can provide access to a machine exhibiting the issue.

erizocosmico:
Fixes #368

### DIFF
--- a/src/main/scala/tech/sourced/engine/iterator/CommitIterator.scala
+++ b/src/main/scala/tech/sourced/engine/iterator/CommitIterator.scala
@@ -98,11 +98,12 @@ object CommitIterator {
       )
     }
 
+    val isChained = ref.isDefined
     val hashKeys = Seq("hash", hashKey)
     var iter: Iterator[ReferenceWithCommit] = new RefWithCommitIterator(
       repo,
       refs,
-      if (!filters.hasFilters("index")) 1 else 0
+      if (isChained && !filters.hasFilters("index")) 1 else 0
     )
 
     if (filters.hasFilters(hashKeys: _*)) {
@@ -147,6 +148,7 @@ private[iterator] class RefWithCommitIterator(repo: Repository,
     while ((commits == null || !commits.hasNext) && refs.hasNext) {
       actualRef = refs.next()
       index = 0
+      consumed = 0
       commits =
         try {
           Git.wrap(repo).log()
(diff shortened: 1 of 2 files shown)


src/main/scala/tech/sourced/engine/iterator/CommitIterator.scala
```diff
@@ -98,11 +98,12 @@ object CommitIterator {
       )
     }
 
+    val isChained = ref.isDefined
     val hashKeys = Seq("hash", hashKey)
     var iter: Iterator[ReferenceWithCommit] = new RefWithCommitIterator(
       repo,
       refs,
-      if (!filters.hasFilters("index")) 1 else 0
+      if (isChained && !filters.hasFilters("index")) 1 else 0
```
smola: Not completely related to this PR, but you might want to consider "index = 0" as optimized with no history traversal.

erizocosmico: Will open an issue about that

PR #380, opened 2018-04-20
provider: make sure all files are cleaned up

### LINKED ISSUE #348: Temporary files are permanent with standalone spark cluster
## Expected Behavior

After a spark job using the engine to process data completes, the engine should remove all temporary fields it created. Ideally, inputs should be removed after each spark task ends.

## Current Behavior
When running jobs on the Kubernetes cluster, for example ml's `repos2ids`, if we use spark in standalone mode then whether a job fails or not, all temporary data created in the worker pods remains. This is a definitive blocker when trying to run jobs on large inputs. This does not happen when running the job with the local spark.

## Possible Solution

When using spark in standalone the engine must do cleaning up on the workers, if possible between spark tasks (if we have to wait for the job to finish when processing PGA for it to start cleaning up, we will need a lot more disk memory)

## Steps to Reproduce (for bugs)

- Connect to cluster and launch a pod:
```
kubectl config use-context local-gpu
kubectl run -i --tty engine-test --image=srcd/ml-core --generator="run-pod/v1" --command -- /bin/bash
```
- Update dependencies, then launch job locally to check the temporary files are collected:
```
pip3 uninstall sourced-engine -y  # 0.5.3 is installed by defaut atm
pip3 install sourced-engine sourced-ml -y
srcml repos2ids -r hdfs://hdfs-namenode/5-siva -o here_out --dep-zip --conf spark.local.dir=/spark-temp-data
du -h /spark-temp-data
```
- Launch job with standalone: `srcml repos2ids -r hdfs://hdfs-namenode/5-siva -o here_out --dep-zip --conf spark.local.dir=/spark-temp-data -s spark://l-spark-spark-master:7077`
- Connect to any worker pod: `kubectl exec -it l-spark-spark-worker-<ID> bash`
-temp files should still be there: `du -h /spark-temp-data`

## Context

This has affected me, as well as all member of the ml team that tried launching jobs using `engine` to process large inputs on `local-gpu` or `staging` (in my case `apollo`).  As a consequence of this issue, temporary data fills up the worker pods, until the Kubernetes controller senses disk pressure on one or multiple nodes, kills the workers and causes our jobs to fail. 
To give a sense of the impact of this, we have not been able to process files larger then 20G on the cluster atm.

For more info, check out problem 2 of [this google doc](https://docs.google.com/document/d/1M2nHwCagndw0h9-i7h_Gwl9kISoOuIX5J_yXjYOuvZI/edit?ts=5ab25fd2).

## Your Environment (for bugs)
* Version used: last, but this was happenning since at least 0.5.2
* Operating System and version: pods on the cluster so Ubuntu 16.04 I believe
* Some needed resources to reproduce the problem: Access to one of the clusters.

erizocosmico:
Fixes #348 

There are a couple changes here:

- If the repository is a siva repository, we don't cleanup the repo directory but it's parent, because siva repositories are created inside another directory, that way, when everything is deleted the only thing left is `processing-repositories` and `siva-files`.
- Adds cleanup to the `TaskContext` on completion and on failure. If for any reason the close did not get called, one of these handlers will cleanup the files. If the files have already been cleaned up, the listeners will do nothing.

With current 0.5.7 I managed to get a few repositories not cleaned up when executing it in standalone mode. I don't know exactly why, because I could not reproduce with custom jar for debugging, but with this for sure won't happen (at least I hope it won't).

I've executed a few runs with a custom jar built with this and no file was left behind, so let's hope this solves ML's problems.

### DIFF
--- a/src/main/scala/tech/sourced/engine/provider/RepositoryProvider.scala
+++ b/src/main/scala/tech/sourced/engine/provider/RepositoryProvider.scala
@@ -8,6 +8,7 @@ import org.apache.commons.pool2.impl.{DefaultPooledObject, GenericKeyedObjectPoo
 import org.apache.commons.pool2.{BaseKeyedPooledObjectFactory, PooledObject}
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem, Path}
+import org.apache.spark.TaskContext
 import org.apache.spark.internal.Logging
 import org.eclipse.jgit.lib.{Repository, RepositoryBuilder}
 import tech.sourced.engine.Sources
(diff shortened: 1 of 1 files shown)


src/main/scala/tech/sourced/engine/provider/RepositoryProvider.scala
```diff
@@ -101,24 +102,61 @@ class RepositoryObjectFactory(val localPath: String, val skipCleanup: Boolean)
...
+    }
+
+    if (!skipCleanup) {
+      // Make really sure that repositories are cleaned up.
+      TaskContext.get().addTaskCompletionListener((_) => {
+        try {
+          cleanupRepo(key, repo)
+        } catch {
+          case e: Throwable =>
+            logDebug(e.getMessage)
```
kuba--: question - debug level is good enough here? just message is good enough

kuba--: maybe we can wrap it in one method and do `scala.util.Try` and `onFailure` log message?

erizocosmico: I think we should actually be fine removing the whole try/catch. I don't think this can throw any exception.

src/main/scala/tech/sourced/engine/provider/RepositoryProvider.scala
```diff
@@ -101,24 +102,61 @@ class RepositoryObjectFactory(val localPath: String, val skipCleanup: Boolean)
...
+        genRepository(conf, path, localPath, isBare = false)
+    }
+
+    if (!skipCleanup) {
+      // Make really sure that repositories are cleaned up.
+      TaskContext.get().addTaskCompletionListener((_) => {
+        try {
+          cleanupRepo(key, repo)
+        } catch {
+          case e: Throwable =>
```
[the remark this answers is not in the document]
erizocosmico: I think I'm just going to remove the try/catch. We're doing a close, which _shouldn't_ throw and a deleteQuietly which does not throw either.

erizocosmico:
Done

ajnavarro:
some errors on CI, @erizocosmico can you have a look?

erizocosmico:
Taking a look at it

erizocosmico:
Fixed

PR #395, opened 2018-05-24
*: add config setting to skip read errors

### LINKED ISSUE #385: [feature request] Handy handling of bad siva-files
following after https://github.com/src-d/engine/issues/377

## Expected Behavior
If there are some bad Siva files or objects, engine should warn about that and skip this file or object. 

## Current Behavior
it crashes with the error like in https://github.com/src-d/engine/issues/377:
```
18/04/18 12:38:55 WARN TaskSetManager: Lost task 143.0 in stage 2.0 (TID 441, 10.2.20.9, executor 0): java.lang.ArrayIndexOutOfBoundsException
	at org.eclipse.jgit.internal.storage.pack.BinaryDelta.apply(BinaryDelta.java:196)
        ...
```

## Why it is important
There are always some unexpected cases in big data. It is really sad when you run something big for several days and It failed because of one small and not so important chunck of data. That is why usually big data systems trying to handle such errors in a soft way, just letting user now, but continue to run. 

## Priority 
Since @src-d/machine-learning team are trying to proccess whole PGA dataset (and it is ok for us to have excluded bad siva-files because of errors) I think it is important feature to have.

### LINKED ISSUE #393: Add a configuration flag to skip siva read errors
To avoid problems reading siva files, add a configuration param to skip errored siva files.

erizocosmico:
Closes #385 
Closes #393


erizocosmico:
Added also to Python bindings. I completely forgot about the Python side 😓

ajnavarro:
LGTM

bzz:
👍 
To save some time for future readers - it's `spark.tech.sourced.engine.skip.read.errors` option

PR #403, opened 2018-06-18
add support for Apache Spark 2.3

smola:
- A new private package `compat` is added with compatibility shims.

- Shims for apply/unapply on LogicalRelation are added with compatibility
  with Spark 2.2.x and 2.3.x.

- If an unsupported Spark version is used, a human-readable exception is thrown.

### DIFF
--- a/src/main/scala/tech/sourced/engine/compat/compat.scala
+++ b/src/main/scala/tech/sourced/engine/compat/compat.scala
@@ -1,0 +1,76 @@
+package tech.sourced.engine.compat
+
+import org.apache.spark.SPARK_VERSION
+import org.apache.spark.sql.catalyst.catalog.CatalogTable
+import org.apache.spark.sql.catalyst.expressions.AttributeReference
+import org.apache.spark.sql.execution.datasources.{
+  LogicalRelation => SparkLogicalRelation
+}
+import org.apache.spark.sql.sources.BaseRelation
+
+import scala.reflect.runtime.{universe => ru}
+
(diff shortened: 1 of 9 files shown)


smola:
Ideally we would add a test build on Travis with Spark 2.3.x.

src/main/scala/tech/sourced/engine/compat/compat.scala
```diff
@@ -0,0 +1,76 @@
...
+}
+import org.apache.spark.sql.sources.BaseRelation
+
+import scala.reflect.runtime.{universe => ru}
+
+private[compat] object Compat {
+
+  def apply[T](s22: T, s23: T): T = SPARK_VERSION match {
+    case s if s.startsWith("2.2.") => s22
+    case s if s.startsWith("2.3.") => s23
```
ajnavarro: Should we use complete versions? I think we needed to change some internal call to move from 2.2.0 to 2.2.1

ajnavarro: nevermind, It was moving from 2.1.x to 2.2.x

smola:
@ajnavarro :+1: Let's not merge this until proper CI is in place.

bzz:
Great! Just sharing, in case this example of [CI from Gemini](https://github.com/src-d/gemini/blob/master/.travis.yml#L64), that can be used with different Spark versions could help here.

smola:
Added CI (migrated to Travis Build Stages).

PR #408, opened 2018-06-26
Fix bblfsh driver versions to avoid github connection problems.

ajnavarro:
Signed-off-by: Antonio Jesus Navarro Perez <antnavper@gmail.com>


erizocosmico:
What does the versions have to do with GitHub? Aren't the driver images on docker hub?

ajnavarro:
The images are on Docker hub, but to check the available drivers, `bblfshd` goes to Github API and check for `-driver` repositories or something like that. Doing this on Travis causes a lot of rate limit errors.

